PR #161, opened 2019-06-25
Add ID to standard props and remove . from displayNames

### LINKED ISSUE #158: Add "id" to standardProps
Would be useful in a few places in the InfluxDB UI. Could use the `testID` prop but for cleanliness `id` is a cleaner way to simply identify a unique element

### LINKED ISSUE #160: Having a . character in a component's displayName prevents storybook from displaying the props table correctly
Screen Shot 2019-06-24 at 5 17 35 PM

alexpaxton:
Closes #160
Closes #158 

### Changes

- Add `id` to `StandardProps` and implement in all components
- Remove `.` from all `displayName`s in components
- Ensure all components are extending `StandardProps`

### Checklist

- [ ] Change is reflected in documentation
- [x] Added entry to top of `changelog.md` with link to PR
- [ ] Signed CLA

### DIFF
--- a/src/Types/index.tsx
+++ b/src/Types/index.tsx
@@ -3,6 +3,8 @@ export interface StandardProps {
   className?: string
   /** ID for Integration Tests */
   testID: string
+  /** Unique identifier for getting an element */
+  id?: string
 }
 
 export enum ComponentColor {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### 0.0.16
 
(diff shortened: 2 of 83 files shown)


src/Types/index.tsx
```diff
@@ -3,6 +3,8 @@ export interface StandardProps {
   className?: string
   /** ID for Integration Tests */
   testID: string
```
chnn: should this be optional?

alexpaxton: It has a default value on all the components

PR #163, opened 2019-06-26
Make Panels Dismissable

### LINKED ISSUE #162: Make panels dismissable
We could achieve this by adding an optional `onDismiss` prop to a component in the `panel` family that would render an X in the corner.

alexpaxton:
Closes #162

### Changes

- Introduce the `onDismiss` prop to `Panel`
  - I added this to the parent element rather than `PanelHeader` so that a panel can be dismissable regardless of its use of a header
- Tweaked the proportions of panels
  - Now they use the global variables `$form-xs-height` (for example)
- Dismiss button changes size in response to panel size prop
- Improved documentation
  - Wrote docs specifically for each example rather than a single example readme
  - Wrote docs for dismissable panels

### Screenshots

dismissable-panel-sizes
Screen Shot 2019-06-26 at 11 51 29 AM

### Checklist
Check all that apply


src/Components/Panel/Panel.scss
```diff
@@ -30,45 +32,114 @@
...
+    opacity: 1;
+  }
+}
+
 /*
   Size Modifiers
   ------------------------------------------------------------------------------
 */
-@mixin panelSizeModifier($fontSize, $fontWeight, $padding) {
+@mixin panelSizeModifier($fontSize, $fontWeight, $padding, $dismissSize) {
```
chnn: From the GIF, it felt like the line height for the `xs` size was a bit large.

alexpaxton: Ah yeah I know watcha mean. Currently the children of `PanelBody` are not affected by the size of `Panel` but I was thinking about having some styles to affect typographic elements within panels to respond to panel sizing

chnn: That would be rad, but nbd either way.

Issue #176, opened 2019-07-18
Port context menu component from InfluxDB


alexpaxton:

daviddkkim:
https://us-east-1-1.aws.cloud2.influxdata.com/orgs/5ad804dc819b2ab5/data-explorer

Right click query1

PR #202, opened 2019-08-01
feat: Introduce Popover Component

### LINKED ISSUE #168: Standardize Popover component
Add a popover component that standardizes the interaction.

- [ ] Tip Text Popover
- [ ] Dismissable Popover

image

image

image

image

image

alexpaxton:
Closes #168

This work is intended for use in building improved versions of `ContextMenu` ( #176 ) and `ConfirmationButton` ( #172 )

### Changes

- Introduce `Popover` component
- Can be customized in appearance and interactions
  - Can be set to `Solid` or `Outline` type
- Doesn't use any new libraries
- Includes optional `<Popover.DismissButton />` that when used, always appears in the top right corner of the popover
- Popovers should always appear within the viewport and on top of everything else
- Should be able to be used for all of the mismatched "popovers" currently in InfluxDB

### Screenshots

Screen Shot 2019-08-01 at 10 22 26 AM
popover-preview
Screen Shot 2019-08-01 at 10 46 34 AM

### Checklist
Check all that apply

- [x] Tests pass


src/Components/Popover/PopoverDismissButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+  StandardProps,
+  IconFont,
+  ComponentSize,
+} from '../../Types'
+
+interface Props extends StandardProps {
+  /** Button color */
+  color: ComponentColor
+  /** Pass the "onHide" function in here */
+  onClick: any
```
alexpaxton: When I typed this as `() => void` I got some weird type errors that I could not figure out how to resolve:

```
TS2322: Type '(() => void) | undefined' is not assignable to type '() => void'.
  Type 'undefined' is not assignable to type '() => void'.
```

src/Components/Popover/Popover.tsx
```diff
@@ -0,0 +1,476 @@
...
+  /** Renders the popover dialog visible initially */
+  initiallyVisible: boolean
+}
+
+interface State {
+  expanded: boolean
+  triggerRect: ClientRect | DOMRect | null
+}
+
+interface PopoverFlush {
```
alexpaxton: Although this is internal to the Popover component, I'm not sure if it's the most understandable name

alexpaxton:
I added a bunch of comments in the Popover component to explain the positioning logic. I had a feeling that I wouldn't remember how it worked if I had to look at it again in a month, so it's for future me and whoever else works on it.

src/Components/Popover/Popover.tsx
```diff
@@ -0,0 +1,476 @@
...
+  last: boolean
+}
+
+export class Popover extends Component<Props, State> {
+  public static readonly displayName = 'Popover'
+
+  public static defaultProps = {
+    color: ComponentColor.Primary,
+    testID: 'popover',
+    showEvent: PopoverInteraction.Click,
```
chnn: I know this isn't the time or place... but I think clockface would be a lot smoother to use if it presented types like

`type PopoverInteraction = 'click' | 'hover'`

instead of

```
enum PopoverInteraction {
  Click = 'click',
  Hover = 'hover',
}
```

src/Components/Popover/Popover.tsx
```diff
@@ -0,0 +1,476 @@
...
+  private get dialog(): JSX.Element {
+    const {triggerRect, expanded} = this.state
+    const {contents, distanceFromTrigger} = this.props
+    const dialogPosition = this.calculateDialogPosition()
+    let dialogStyles: CSSProperties = {
+      visibility: expanded ? 'visible' : 'hidden',
+    }
+    let caretStyles: CSSProperties = {}
+
+    if (triggerRect && this.dialogRef.current) {
```
chnn: It seems like this needs `dialogRef.current` to render, but `dialogRef.current` isn't attached until it renders. So it requires two renders to work, which may be happening incidentally. But it's hard to tell. Will this work if it is rendered exactly once?

One other way to do it is to render the dialog with its contents (which attaches the ref), but then to compute the styles in `componentDidUpdate` and attach them imperatively (i.e. `this.dialogRef.current.setAtribute('style', ...)`).

src/Components/Popover/Popover.tsx
```diff
@@ -0,0 +1,476 @@
...
+      const dialogRect = this.dialogRef.current.getBoundingClientRect()
+      let dialogFlush
+
+      switch (dialogPosition) {
+        case PopoverPosition.Above:
```
chnn: Might be helpful to extract this to a pure function or composition of pure functions. This logic is gnarly and I know it has to be, but maybe we can prevent it from become gnarlier by explicitly declaring the dependencies of this routine as the arguments to a function.

Otherwise it can access everything on `this`, which is far more than it needs. I think over time that leads to tight coupling / people trying to do too much all in one place.

src/Components/Popover/Popover.tsx
```diff
@@ -0,0 +1,476 @@
...
+        className="cf-popover--dialog"
+        style={dialogStyles}
+        ref={this.dialogRef}
+      >
+        <div className="cf-popover--dialog-contents">
+          <div
+            className={`cf-popover--caret cf-popover--caret__${dialogPosition}`}
+            style={caretStyles}
+          />
+          {contents(this.handleHideDialog)}
```
chnn: This render props looks really great to use.

src/Components/Popover/Popover.tsx
```diff
@@ -0,0 +1,476 @@
...
+    this.setState({
+      triggerRect: this.triggerRef.current.getBoundingClientRect(),
+    })
+  }
+
+  private handleShowDialog = (): void => {
+    this.handleUpdateTriggerRect()
+
+    window.addEventListener('resize', this.handleUpdateTriggerRect)
+    window.addEventListener('scroll', this.handleUpdateTriggerRect)
```
chnn: I think this will miss the case of the element moving on screen even if the window doesn't resize or the user doesn't scroll. For example if a sidebar is closed, everything on the side of it will shift. Kinda a minor concern that seems hard to fix.

src/Components/Popover/Popover.tsx
```diff
@@ -0,0 +1,476 @@
...
+
+    this.setState({expanded: true})
+  }
+
+  private handleHideDialog = (): void => {
+    this.setState({expanded: false})
+
+    window.removeEventListener('resize', this.handleUpdateTriggerRect)
```
chnn: Will this be fired when the component unmounts? It looks like it relies on a `mouseOut` event firing, but I'm not sure that will happen. Could be leaking memory.

src/Components/Popover/Popover.tsx
chnn: A comment with an example might help.

src/Components/Popover/Popover.tsx
alexpaxton: Good catch! I’ll add that in as a precaution

src/Components/Popover/Popover.tsx
alexpaxton: Ah that's a good point. I have no idea how do fix that unless there was some sort of fancy `componentDidMove` event

src/Components/Popover/PopoverDismissButton.tsx
mavarius: You'd need to type it as `() => void) | undefined`

src/Components/Popover/Popover.md
```diff
@@ -0,0 +1,77 @@
...
+    </div>
+  )}
+>
+  <div>Trigger Element</div>
+</Popover>
+```
+
+### Gotchas
+
+The popover dialog makes use of `position: fixed` and `z-index: 9999` to ensure it is never obscured. In order to keep the dialog positioned next to the trigger the component listens for scroll and resize events while the popover is visible. This causes the dialog to re-render more often, which may be troublesome depending on what kind of component you render as the contents.
```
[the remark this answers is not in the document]
alexpaxton: That actually sounds a tad cleaner. Wouldn't be radically different since fixed positioning is relative to the viewport, so we wouldn't have to change the CSS during the change

PR #271, opened 2019-09-11
First pass at a contributing guide

alexpaxton:
the contributing guide has some basics and will likely need more work, but having an incomplete guide is better than no guide

I mainly focused on documenting our code patterns

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,164 @@
+# Want to Contribute to Clockface?
+
+First things first sign our [CLA](https://influxdata.com/community/cla/)
+
+Clockface is built using `React`, `Typescript`, `SCSS`, and `Storybook` so you will want to get familar with those tools.
+We use `yarn` to manage dependencies and build the project.
+
+Get to know our key maintainers:
+
+@thealexpaxton
+@mavarius
+
+## Submitting a Pull Request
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,164 @@
+# Want to Contribute to Clockface?
+
+First things first sign our [CLA](https://influxdata.com/community/cla/)
+
+Clockface is built using `React`, `Typescript`, `SCSS`, and `Storybook` so you will want to get familar with those tools.
+We use `yarn` to manage dependencies and build the project.
```
mavarius: and `Webpack`

CONTRIBUTING.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+### 3. The Component Family Pattern
+
+We offer larger groups of components that are intended to be used together closely. In these cases we consider those components a "family". Component families not only share a namespace but are accessible via a single import. This is the only time we use classes.
+
+##### Here's an example:
+
+```tsx
+import {Apple} from './Apple'
```
mavarius: Imports here should be `FruitApple` and `FruitOrange`

alexpaxton: Ah good catch

PR #327, opened 2019-10-11
fix: popover visibility

### LINKED ISSUE #326: Popovers that are initially visible stuck in top left
**Describe the bug**
It appears that when a `Popover` is set to be visible initially it renders in the top left corner and is unaffected by scroll or resize events

**To Reproduce**
I tested `master` of Clockface within InfluxDB using a relative path in `package.json`
Visit the `New Threshold Check` or `New Deadman Check` overlay
Notice the popover is not attached to the "Save" button

**Expected behavior**
The `Popover` should be affixed to the save button in the top right, and reposition when the screen is resized or scrolled

**Screenshots**
Screen Shot 2019-10-11 at 11 05 31 AM

alexpaxton:
Closes #326 

### Changes

- Remove `<span />` wrapping chil in `Popover`
- Popover `ref` prop passes through to "contents" element within `PopoverDialog`
- Use `useRef` instead of `createRef`
- Add popover to storybook examples that is visible by default (for testing purposes)
- Extract `createPortal` and `destroyPortal` functions into `/utils`
- Fire `handleUpdateStyles` on `PopoverDialog` initial render

Tested all changes in `influxdata/influxdb`

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/src/Components/Popover/Base/Popover.tsx
+++ b/src/Components/Popover/Base/Popover.tsx
@@ -3,8 +3,8 @@ import React, {
   useState,
   useEffect,
   RefObject,
-  forwardRef,
   MouseEvent,
+  forwardRef,
 } from 'react'
 import {createPortal} from 'react-dom'
 import uuid from 'uuid'
@@ -26,6 +26,7 @@ import {
   PopoverInteraction,
   StandardFunctionProps,
 } from '../../../Types'
+import {PopoverDialogRef} from './PopoverDialog'
 
 export interface PopoverProps extends StandardFunctionProps {
   /** Popover dialog color */
@@ -58,7 +59,7 @@ export interface PopoverProps extends StandardFunctionProps {
   enableDefaultStyles?: boolean
 }
 
-export type PopoverRef = HTMLSpanElement
+export type PopoverRef = PopoverDialogRef
 
 interface CustomMouseEvent extends MouseEvent {
   relatedTarget: Element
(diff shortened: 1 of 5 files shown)


src/Components/Popover/Base/Popover.tsx
```diff
@@ -55,13 +59,13 @@ export interface PopoverProps extends StandardFunctionProps {
...
 
-export type PopoverRef = HTMLSpanElement
+export type PopoverRef = PopoverDialogRef
 
 interface CustomMouseEvent extends MouseEvent {
   relatedTarget: Element
 }
 
-let uniquePortalID: string = ''
+const popoverPortalName = 'popover'
```
mavarius: Is this to be consistent for all shared for all popovers globally?

src/Components/Popover/Base/Popover.tsx
```diff
@@ -88,6 +92,28 @@ export const PopoverRoot = forwardRef<PopoverRef, PopoverProps>(
     ref
   ) => {
     const [expanded, setExpanded] = useState<boolean>(!!visible)
+    const [portalID, setPortalID] = useState<string>('')
+
+    useEffect((): (() => void) => {
+      const newPortalID = `cf-${popoverPortalName}-portal-${uuid.v4()}`
+      !portalID && setPortalID(newPortalID)
```
mavarius: Since this effect only happens once, is this check actually doing anything?

src/Components/Popover/Base/Popover.tsx
alexpaxton: yarrrr

src/Components/Popover/Base/Popover.tsx
alexpaxton: I suppose it might be excessive, I'll test it w/o the check and see what happens

PR #340, opened 2019-10-16
Fix Validated Components

### LINKED ISSUE #339: Color picker always invalid
**Describe the bug**
ColorPicker is always passing back its new status as `Error` even when a valid hexcode is in there

This is causing some e2e tests to fail in InfluxDB

alexpaxton:
Closes #339 

Turns out the bug was caused by the hexcode validator function always returning a string when it was expecting a `null` value to determine if is valid, hence it always returning invalid

### Changes

- Change typing of `FormValidationElement` to expect `string` instead of `any`
  - Turns out the component is exclusively used to validate text inputs
- Remove state from `ColorPicker` and `FormValidationElement`
  - Should make these work like controlled components
- Create a type for `ValidationFunction` to enforce consistency across these components (and potentially more)
  - Validation functions should return a string (error) or null (valid)
- Add `validationFunc` prop to `ColorPicker`
  - Has a hexcode validator by default
  - User can pass in their own function
- Updated stories to include inputs for testing purposes

### Checklist
Check all that apply

- [x] Tests pass


src/Components/Button/Composed/ConfirmationButton.tsx
```diff
@@ -110,9 +111,9 @@ const ConfirmationContents: FunctionComponent<{
   returnValue?: any
   onConfirm: (returnValue?: any) => void
   confirmationLabel: string
-  testID?: string
+  testID: string
```
drdelambre: 😲

src/Components/ColorPicker/ColorPicker.tsx
```diff
@@ -60,35 +57,26 @@ export const ColorPicker = forwardRef<ColorPickerSwatchRef, ColorPickerProps>(
       colors = influxColors,
       testID = 'color-picker',
       maintainInputFocus = false,
+      validationFunc = validateHexCode,
     },
     ref
   ) => {
-    const [errorMessage, setErrorMessage] = useState('')
+    const errorMessage = validationFunc(color)
```
drdelambre: shouldn't this be more of a `validateHexCode(color) && validationFunc(color)` relationship? so that validating that a color matches a palette map somewhere doesnt mean you have to rewrite validateHexCode?

src/Components/ColorPicker/ColorPicker.tsx
```diff
@@ -60,35 +57,26 @@ export const ColorPicker = forwardRef<ColorPickerSwatchRef, ColorPickerProps>(
...
-      const cleanedValue = trimmedValue
-        .split('')
-        .filter(char => HEX_CHARACTERS.includes(char.toLowerCase()))
-        .join('')
-
-      const newErrorMessage = validateHexCode(cleanedValue)
-
-      setErrorMessage(newErrorMessage)
-      onChange(cleanedValue, inputStatus)
+      onChange(e.target.value, inputStatus)
```
drdelambre: why no more input cleaning here?

src/Components/ColorPicker/ColorPicker.tsx
alexpaxton: I made `validationFunc` to be `validateHexCode` by default, but in case we ever want to support rgb values or named colors a different validation function can be passed in

src/Components/ColorPicker/ColorPicker.tsx
alexpaxton: I suppose I could trim out space characters at the very least

alexpaxton:
@drdelambre can you clarify "Chaining of validation"?

drdelambre:
like.. whats the difference between a text input, color hex input, and a "only shades of grey" input? is the thinking that the last input is going to be `custom validation function + hex validation function + text validation function`, or is it `custom validation function implements (hex validation function + text validation function)`? is the idea that v1.0 is a transition from the former to the later?

src/Components/ColorPicker/ColorPicker.tsx
drdelambre: 👍

PR #368, opened 2019-11-01
Fix Popover hide bug

### LINKED ISSUE #367: Always visible Popover hidden if trigger element initially renders off screen
The recently added feature which hides `PopoverDialog` if the trigger element is not visible is colliding with the "always visible" feature. The logic here
https://github.com/influxdata/clockface/blob/master/src/Components/Popover/Base/PopoverDialog.tsx#L112-L118

Should take into account the `visible` prop

alexpaxton:
Closes #367 

### Changes

- Pass `visible` prop into `PopoverDialog` and prevent the intersection hide from firing if `visible` is true

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/src/Components/Popover/Base/PopoverDialog.tsx
+++ b/src/Components/Popover/Base/PopoverDialog.tsx
@@ -46,6 +46,8 @@ export interface PopoverDialogProps extends StandardFunctionProps {
   enableDefaultStyles: boolean
   /** Allows the popover to dismiss itself when the trigger is no longer in view */
   onHide: () => void
+  /** This keeps the Popover visible no matter what */
+  visible?: boolean
 }
 
 export type PopoverDialogRef = HTMLDivElement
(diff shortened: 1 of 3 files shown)


src/Components/Popover/Base/PopoverDialog.tsx
```diff
@@ -46,6 +46,8 @@ export interface PopoverDialogProps extends StandardFunctionProps {
   enableDefaultStyles: boolean
   /** Allows the popover to dismiss itself when the trigger is no longer in view */
   onHide: () => void
+  /** This keeps the Popover visible no matter what */
+  visible?: boolean
```
hoorayimhelping: does `visible` default to `true` or `false`? if it defaults to false, it's probably fine to leave this an optional param. if it should default to `true`, we should probably make it required and provide a default boolean value of `true` on line 64 below

alexpaxton: It is false by default, but it wouldn’t hurt to have that more explicitly stated

PR #377, opened 2019-11-08
Introduce RightClick component

### LINKED ISSUE #171: Port right click menu from InfluxDB

alexpaxton:
Closes #171 

Although this component looks and works similarly to `Popover`, there are some key differences:

- `RightClick` is positioned wherever the right click happened, rather than being strictly anchored to the trigger element's position
- `RightClick` does not have a "caret" or customizable `position` so the logic is much simpler

### Changes

- Introduce a component that works identically to `RightClick` as defined in InfluxDB
- Make use of `createPortal` / `destroyPortal` as used in `Popover` and `Overlay`
- Extend the component to come in different colors
- Add some rudimentary positioning logic so the menu will avoid appearing off screen

### Screenshots

right-click-2

### Checklist
Check all that apply

- [x] Tests pass


src/Components/RightClick/Base/RightClick.scss
```diff
@@ -0,0 +1,98 @@
+@import '../../../Styles/modules';
```
drdelambre: nit: new to clockface, but why aren't we using absolute positioning to make migrations and restructuring of folders easier during maintenance mode on the code base?

src/Components/RightClick/Base/RightClick.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+      return (): void => {
+        destroyPortalElement(newPortalID)
+        handleRemoveEventListeners()
+      }
+    }, [])
+
+    const handleTriggerClick = (e: MouseEvent): void => {
+      if (disabled || !triggerRef.current) {
+        // triggerRef.current can be set to null in a specific context:
```
drdelambre: ⭐️

src/Components/RightClick/Base/RightClick.tsx
```diff
@@ -0,0 +1,153 @@
...
+    const mouseOffset = useRef<Coordinates>({x: 0, y: 0})
+
+    useEffect((): (() => void) => {
+      const newPortalID = `cf-${rightClickPortalName}-portal-${uuid.v4()}`
+      setPortalID(newPortalID)
+
+      createPortalElement(newPortalID, rightClickPortalName)
+      handleAddEventListeners()
+
+      return (): void => {
```
drdelambre: nit: you should normally reverse the order of your destructors to defensively manage side effects that might sneak in, but not a big deal.

src/Components/RightClick/Base/RightClick.tsx
```diff
@@ -0,0 +1,153 @@
...
+    const handleAddEventListeners = (): void => {
+      if (!triggerRef.current) {
+        return
+      }
+
+      triggerRef.current.addEventListener('contextmenu', handleTriggerClick)
+    }
+
+    const handleRemoveEventListeners = (): void => {
+      if (!triggerRef.current) {
```
drdelambre: this should never resolve to true right?

src/Components/RightClick/Base/RightClickMenu.tsx
```diff
@@ -0,0 +1,123 @@
...
+        onHide()
+      }
+    }
+
+    const observer = new IntersectionObserver(hidePopoverWhenOutOfView)
+
+    useLayoutEffect((): (() => void) => {
+      handleUpdateStyles()
+      observer.observe(triggerRef.current)
+      // The third argument in addEventListener is "false" by default and controls bubbling
```
drdelambre: 😆 and no one remembers why until this moment

src/Components/RightClick/Base/RightClick.scss
alexpaxton: We tried to set it up and failed, so it's on the back burner. If y'all can make it work we are very down

src/Components/RightClick/Base/RightClickMenu.tsx
alexpaxton: Helpful comment is helpful

src/Components/RightClick/Base/RightClick.tsx
alexpaxton: Yeah, I just have it in there to make typescript happy. Otherwise it types as `RefObject | undefined` and complains

PR #436, opened 2020-02-21
feat: export variables.scss as a plain scss file for import

hoorayimhelping:
Closes #

### Changes

Exports ours variables.scss file as a plain scss file so anything that depends on clockface can access its variables.

this would allow us to do something like the following in Influxdb or giraffe
```scss
@import "reset";
@import "@influxdata/clockface/dist/variables.scss";
@import "mixins";
```

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
 # Changelog
 
-#### 1.1.6 (Unreleased)
+#### 1.2.0
+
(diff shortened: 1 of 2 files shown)


alexpaxton:
@hoorayimhelping just throw an entry in the changelog and you're good! Might have to add a header for the newest version if there isn't one

PR #437, opened 2020-02-27
Pendulum Theme

alexpaxton:
Introducing a comprehensive visual refresh for the library. Also did some housekeeping

##### Cool New Stuff:

- New color palette which incorporates the Megenta & Chartreuse colors
- Remove `Roboto` font and make `Rubik` the primary font
- Remove `RobotoMono` font and replace with `IBMPlexMono`
  - Both these changes reduce the overall size of the binary
- Rename `PageHeader` to `PageControlBar`
- Introduce larger `PageHeader` component (2 row header design)
- Introduce new `TreeNav` component
- Introduce new `Heading` typography component
- Add a handful of icons to the icon font
- Redesigned `Button` components to use gradients
- Redesigned `Tabs` components to be flatter and more minimal

##### Housekeeping:

- Move all variables and mixins into the single `variables.scss` which is now exported
- Standardize all font weights to match what is available in the new fonts
  - Create variables for these to make it easier to maintain
- Move variables from some component stylesheets into `variables.scss`

##### Testing:

I've developed this changeset by creating builds of Clockface and having InfluxDB point to my local build as a dependency instead of the NPM package. I've tested out all the new components and changes in the intended code base.

### Screenshots

Screen Shot 2020-02-25 at 11 37 30 AM
Screen Shot 2020-02-24 at 3 08 40 PM
Screen Shot 2020-02-24 at 10 57 00 AM
Screen Shot 2020-02-24 at 10 56 50 AM

### Checklist
Check all that apply

- [x] Tests pass


src/Components/Popover/Base/Popover.scss
```diff
@@ -122,7 +122,7 @@
 
   strong {
     color: $g18-cloud;
-    font-weight: 900;
+    font-weight: 500;
```
Asalem1: I'm a little confused about this change. In my mind, the `strong` tag is synonymous with `bold`. Rather than overwriting the default to normalize the font-weight, wouldn't we just not want to use `strong`?

src/Components/Table/Table.scss
```diff
@@ -21,7 +21,7 @@ $table-row-background: $g3-castle;
 
 .cf-table--header-cell {
   color: $g15-platinum;
-  font-weight: 900;  
+  font-weight: 500;
```
Asalem1: Similar to what I mentioned about the transition from `700` to `500`. The leap from `900` to `500` seems like a drastic stylistic change. If the original intention of the font-weight was to have a bold appearance, shouldn't we maintain that here?

src/Components/Tabs/Documentation/Tabs.stories.tsx
```diff
@@ -42,18 +38,16 @@ tabsStories.add(
       /* eslint-enable */
     }
 
+    const handleClick = (id: string): void => {
```
Asalem1: nit: can we update this to `handleTabClick` just to be consistent?

src/Components/TreeNav/TreeNavSubItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+      className={treeNavSubItemClass}
+      style={style}
+      data-testid={testID}
+    >
+      {labelElement}
+    </div>
+  )
+}
+
+TreeNavSubItem.displayName = 'TreeNavSubItem'
```
Asalem1: Should we allow for `icons` to also be passed in to the `TreeNavSubItem`?

src/Styles/fonts.scss
```diff
@@ -1,42 +1,36 @@
...
-  src: url('./Fonts/Roboto-Black.ttf') format('ttf');
+  src: url('./Fonts/Rubik-Bold.ttf') format('ttf');
 }
 @font-face {
-  font-family: 'RobotoMono';
+  font-family: 'Rubik';
   font-style: normal;
-  font-weight: 500;
-  src: url('./Fonts/RobotoMono-Medium.ttf') format('ttf');
+  font-weight: 900;
```
Asalem1: This change seems incongruent. If we're maintaining a `normal` style, shouldn't we expect the weight to match? (i.e. 500, and not 900)?

.storybook/Story.scss
```diff
@@ -414,7 +414,8 @@ body.sb-show-main {
   }
 
   code {
-    font-family: 'RobotoMono', monospace;
+    font-family: 'IBMPlexMono', monospace;
+    font-style: italic;
```
mavarius: Perhaps the default code style shouldn’t be italic.

src/Styles/variables.scss
```diff
@@ -69,14 +116,21 @@ $cf-radius: 4px;
...
+$cf-header-font: 'Rubik', Helvetica, Arial, Tahoma, Verdana, sans-serif;
+$cf-text-font: 'Rubik', Helvetica, Arial, Tahoma, Verdana, sans-serif;
+$cf-code-font: 'IBMPlexMono', monospace;
 $cf-icon-font: 'icomoon';
 
+$cf-font-weight--light: 300;
+$cf-font-weight--regular: 400;
+$cf-font-weight--medium: 500;
+$cf-font-weight--bold: 700;
+$cf-font-weight--black: 900;
```
alexpaxton: @Asalem1 check it

src/Components/Dropdowns/Dropdown.scss
```diff
@@ -35,7 +35,7 @@ $dropdown-item--padding-v: $cf-marg-a + $cf-border;
   position: absolute;
   width: 100%;
   left: 0;
-  z-index: 500;
+  z-index: 5000;
```
mavarius: If we make variables of the z-indexes across the application it may be easier to avoid layering issues.

src/Components/Form/Form.scss
```diff
@@ -100,7 +100,7 @@
 .cf-form--box {
   width: 100%;
   border-radius: $cf-radius;
-  background-color: $g4-onyx;
-  padding: $cf-marg-b 11px;
+  background-color: $cf-card-background--default;
+  padding: $cf-marg-b ($cf-marg-b + $cf-marg-a);
```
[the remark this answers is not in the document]
mavarius: Ditto this. We seem to be using this calculation in other areas as well (sometimes reversed as a+b), is there a reason it isn’t one of the defined sizes?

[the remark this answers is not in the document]
alexpaxton: https://github.com/influxdata/clockface/issues/438

src/Components/Logo/InfluxDBCloudLogo.tsx
```diff
@@ -0,0 +1,104 @@
...
+import './InfluxDBCloudLogo.scss'
+
+export interface InfluxDBCloudLogoProps extends StandardFunctionProps {
+  /** Coloration of the SVG image */
+  fill?: InfluxColors | string
+  /** Controls rendering of optional "cloud" suffix */
+  cloud: boolean
+}
+
+export type InfluxDBCloudLogoRef = SVGSVGElement
```
mavarius: SVGElement

src/Components/Logo/InfluxDataLogo.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+// Styles
+import './InfluxDataLogo.scss'
+
+export interface InfluxDataLogoProps extends StandardFunctionProps {
+  /** Coloration of the SVG image */
+  fill?: InfluxColors | string
+}
+
+export type InfluxDataLogoRef = SVGSVGElement
```
mavarius: SVGElement

src/Components/Logo/InfluxDataLogo.tsx
```diff
@@ -0,0 +1,119 @@
...
+      <svg
+        version="1.1"
+        ref={ref}
+        id={id}
+        data-testid={testID}
+        style={logoStyle}
+        x="0px"
+        y="0px"
+        viewBox="0 0 1080 300"
+        width="1080"
```
mavarius: Should size here be controlled by a prop as well? Also, we should lock the aspect ratio so the logo can’t be distorted.

src/Styles/fonts.scss
```diff
@@ -1,42 +1,36 @@
...
+  src: url('./Fonts/Rubik-Black.ttf') format('ttf');
 }
+@font-face {
+  font-family: 'IBMPlexMono';
+  font-style: italic;
```
mavarius: Another note on using italic for the monopsace font.

src/Components/Logo/InfluxDataLogo.tsx
alexpaxton: If I use `SVGElement` it requires `href` and 45 other attributes, and doesn't like `ref`. For some reason `SVGSVGElement` seems to be the correct type for inline svgs + refs

src/Components/Logo/InfluxDataLogo.tsx
alexpaxton: These control the internal proportions of the image. If you pass in a style prop with either `width` or `height` it proportionally scales

src/Components/Form/Form.scss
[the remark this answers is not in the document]
alexpaxton: @mavarius I suppose we could add it as a half-step between `a` and `b`, maybe as `$cf-marg-ab`?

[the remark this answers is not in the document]
mavarius: If our margins and paddings tend to differ, perhaps what we need is a set of `cf-pad-` variables.

PR #465, opened 2020-03-19
Fix gradient mixin

### LINKED ISSUE #459: Gradient mixin preventing transparency
Recently the gradient mixins changed to explicitly set `background-image` / `background-color` instead of relying on the shorthand `background`. This is preventing any gradients that make use of transparency from appearing transparent.

Screen Shot 2020-03-18 at 3 20 16 PM

alexpaxton:
Closes #459

### Changes

- Revert gradient mixins to set only the `background` property in shorthand
- Update contrast tester widget

### Screenshots

Screen Shot 2020-03-19 at 4 46 38 PM

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/src/Utils/colors.ts
+++ b/src/Utils/colors.ts
@@ -1,5 +1,6 @@
 // Libraries
 import {CSSProperties} from 'react'
+import chroma from 'chroma-js'
 
 // Types
 import {Gradients, Gradient, DropdownMenuTheme} from '../Types'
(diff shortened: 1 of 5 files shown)


src/Utils/colors.ts
```diff
@@ -32,3 +33,24 @@ export const generateInlineCSSGradient = (
...
 }
+
+export const getAverageColorFromLinearGradient = (
+  linearGradient: string
+): string => {
+  const rgbColors = linearGradient.match(
+    /[rR][gG][bB][(]\d+[,]\s\d+[,]\s\d+[)]/g
+  )
+
+  // This function could fail if a linear gradient string is passed in using hexcodes
```
mavarius: We can do matching on the hex as well.

PR #482, opened 2020-04-13
Refactor ResourceCard

### LINKED ISSUE #479: Do not wrap resource card items in div
I am currently trying to customize the layout of resource cards a bit and am fighting the current implementation. I am making smaller width cards in which I need different pieces to fill space differently. It is difficult to control because every card item is wrapped in `div.cf-resource-card--row`, which doesn't seem to be adding much benefit

alexpaxton:
Closes #479

### Problem

- The design and styling of resource cards is opinionated and inflexible, making them hard to customize
- The render props based interface limits control
  - Can't control ordering of children
- Every render prop getting wrapped is excessive and makes it hard to style children

### Changes

- Remove all render props from `ResourceCard` and offer `children` as the one way to pass things in
- Make `ResourceCardMeta` its own component
- Refactor `ResourceCard` to be an extension of `FlexBox`
  - Reduces custom styles
  - Allows the orientation, spacing, and alignment of the cards to be customizable
- Make the highlight on hover interaction optional
- Make context menu visibility controllable

### Screenshots

resource-card

### Checklist

Check all that apply

- [x] Tests pass


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 #### 2.0.4 (Unreleased)
 
+- [#482](https://github.com/influxdata/clockface/pull/482): [Breaking] Remove render props from `ResourceCard` in favor of `children`
```
hoorayimhelping, line 6: >[Breaking]

this implies that this should be 3.0.0, right? if we're breaking the interface, it's a major breaking change, which is a major bump in semver. What do you think?

src/Components/ResourceList/Card/ResourceCardMeta.tsx
```diff
@@ -0,0 +1,72 @@
...
+      alignItems = AlignItems.Center,
+      justifyContent,
+    },
+    ref
+  ) => {
+    const resourceCardMetaClass = classnames('cf-resource-meta', {
+      [`${className}`]: className,
+    })
+
+    let wrappedChildren = children
```
hoorayimhelping, line 43:
```suggestion
    let wrappedChildren
```

a little more explicit and less confusing. we only want to use `children` if there are children, which is checked on the line below. this way is effectively the same thing, but doesn't require guessing what the value of `children` is.

src/Components/ResourceList/Card/ResourceCardMeta.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    if (React.Children.count(children) > 0) {
+      const childArray: JSX.Element[] = []
+
+      React.Children.forEach(children, child => {
+        childArray.push(<div className="cf-resource-meta--item">{child}</div>)
+      })
```
hoorayimhelping, line 50: style suggestion, doesn't need to be taken. but the kids apparently like `map` better than `forEach` these days.

```suggestion
      const childArray: JSX.Element[] = React.Children.map(children, child => {
        return(<div className="cf-resource-meta--item">{child}</div>)
      })
```

PR #489, opened 2020-04-15
fix(clockface): remove same color used by text from the background in IndexList body

### LINKED ISSUE #488: IndexListBody is bright grey when empty
task-background

TCL735:
Closes https://github.com/influxdata/influxdb/issues/17635
Closes #488 

### Changes

Remove the background color which is the same value as the text color. This applies to IndexList body when it is empty.

### Screenshots

BEFORE: With text and background color the same
Screen Shot 2020-04-15 at 2 49 13 PM

AFTER: Correct background color and text color
Screen Shot 2020-04-15 at 3 08 38 PM

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/src/Components/IndexList/IndexList.scss
+++ b/src/Components/IndexList/IndexList.scss
@@ -154,7 +154,6 @@ $cf-index-list--sort-text: $g11-sidewalk;
 
 // Empty state
 .cf-index-list--empty-cell {
-  background-color: $cf-card-text--disabled;
   border-radius: $cf-radius;
   display: flex;
   align-content: center;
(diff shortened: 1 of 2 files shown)


src/Components/IndexList/IndexList.scss
```diff
@@ -154,7 +154,6 @@ $cf-index-list--sort-text: $g11-sidewalk;
 
 // Empty state
 .cf-index-list--empty-cell {
-  background-color: $cf-card-text--disabled;
```
mavarius, line 157: I think this was supposed to be `$cf-card-background--disabled`

alexpaxton:
MASSIVE

src/Components/IndexList/IndexList.scss
alexpaxton: @mavarius good catch

TCL735: Thanks guys! I have fixed it. Screenshot updated as well.

PR #495, opened 2020-04-30
Add optional "disabled" prop to DropdownItem

### LINKED ISSUE #490: Allow dropdown items to be disabled
Currently there is no way to disable a dropdown item

alexpaxton:
Closes #490 

### Changes

- Allow both `DropdownItem` and `DropdownLinkItem` to be disabled
- Update documentation for `DropdownMenu` to make it easier to test the above changes

### Screenshots

dropdown-item-disabled

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/src/Components/Dropdowns/DropdownItem.tsx
+++ b/src/Components/Dropdowns/DropdownItem.tsx
@@ -21,6 +21,8 @@ export interface DropdownItemProps extends StandardFunctionProps {
   wrapText?: boolean
   /** Title attribute */
   title?: string
+  /** Prevents any interaction with this element, including the onClick function */
+  disabled?: boolean
 }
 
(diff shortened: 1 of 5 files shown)


src/Components/Dropdowns/DropdownItem.tsx
```diff
@@ -21,6 +21,8 @@ export interface DropdownItemProps extends StandardFunctionProps {
   wrapText?: boolean
   /** Title attribute */
   title?: string
+  /** Prevents any interaction with this element, including the onClick function */
+  disabled?: boolean
```
mavarius, line 25: Should we use `status` & `ComponentStatus.Disabled` here like we do in our other components? This would also allow us to disable it during some loading state and we could use it to toggle to active state styles.

PR #502, opened 2020-05-05
Allow popovers to be closed via the Escape key

### LINKED ISSUE #501: Add keyboard interaction to close popover
**Is your feature request related to a problem? Please describe.**
When a popover is opened, currently the only way to exit is by clicking outside of the element via mouse interaction. This is an accessibility issue and also leads to bugs.

**Describe the solution you'd like**
AS a user with a popover open
WHEN I press the `ESC` key
THEN the popover should close

**Additional context**
Eg: An interaction inside the popover opens a dialog but the popover persists.
image

alexpaxton:
Closes #501 

### Changes

- Popovers can be dismissed by hitting the escape key
  - The dialog component can be focusable and receive keyboard events
  - Keyboard events bubble up to the portal containing the popover and then it closes
  - `PopoverDialog` will make itself the browser's active element if nothing else is in focus
- Update documentation
  - Explain nuances of prop combinations
  - Add more permutations of props for manual testing
- Round numbers down to whole numbers when calculating popover position
  - This prevents fuzzy sub-pixel rendering of popovers
- If `visible` is true it will supersede the escape key behavior in the interest of complete external control

### Screenshots

Screen Shot 2020-05-21 at 12 07 48 PM

Example of `PopoverDialog` deferring focus to an input with `autofocus={true}`:
popover-doesnt-pull-focus

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/src/Components/Popover/Base/Popover.tsx
+++ b/src/Components/Popover/Base/Popover.tsx
@@ -53,7 +53,7 @@ export interface PopoverProps extends StandardFunctionProps {
   position?: PopoverPosition
   /** Means of applying color to popover */
   appearance: Appearance
-  /** Overrides internal popover expanded state */
+  /** For external control of the popover state, overrides escape key behavior  */
   visible?: boolean
   /** Disables the popover's show interaction */
   disabled?: boolean
@@ -61,8 +61,6 @@ export interface PopoverProps extends StandardFunctionProps {
   triggerRef: RefObject<any>
   /** Adds reasonable styles to popover dialog contents so you do not have to */
   enableDefaultStyles?: boolean
-  /** If true the popover can be dismissed via the Escape key */
-  enableEscapeKey?: boolean
 }
 
 export type PopoverRef = PopoverDialogRef
@@ -84,13 +82,12 @@ export const PopoverRoot = forwardRef<PopoverRef, PopoverProps>(
       className,
       triggerRef,
       caretSize = 8,
-      visible = false,
+      visible,
       disabled = false,
       testID = 'popover',
       distanceFromTrigger = 4,
       appearance = Appearance.Outline,
       enableDefaultStyles = true,
-      enableEscapeKey = true,
       color = ComponentColor.Primary,
       position = PopoverPosition.Below,
       showEvent = PopoverInteraction.Click,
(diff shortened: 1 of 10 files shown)


src/Components/Popover/Documentation/Popover.md
```diff
@@ -43,13 +43,19 @@ Make sure to use the available `onHide` argument and pass it into your dismiss b
...
 By default `Popover` will apply reasonable default styles to the dialog to save you time. However you may be building a more specialized design and want more control. Rather than fight the default styles you can simply turn them off:
 ```tsx
 <Popover enableDefaultStyles={false} />
 ```
 
+### Gotchas
+
+There is a hierarchy of importance for handling combinations of props in `Popover`. If a `Popover` has its `visible` prop set to `true` then it will be visible so long as that value is unchanged. The dialog will ignore `enableEscapeKey` and defer to the value of `visible`. This decision assumes the intent of `visible` does not want to be disrupted by the escape key interaction.
+
+In addition, if  the Escape key is pressed any and all `Popovers` that can be closed via the Escape key will be closed.
```
mavarius, line 57: The escape key should only close the focused popover. Having it close all popovers at once will lead to unexpected behavior.

src/Components/Popover/Base/Popover.tsx
```diff
@@ -179,7 +205,15 @@ export const PopoverRoot = forwardRef<PopoverRef, PopoverProps>(
       setExpanded(false)
     }
 
-    const handleAddEventListeners = (): void => {
+    const handleAddEscapeKeyListener = (): void => {
+      window.addEventListener('keydown', handleEscapeKey)
```
mavarius, line 209: These listeners should target the popover that initiated it.

alexpaxton:
@mavarius I'm gonna noodle on this more and see if I can find a way to only close a single popover at a time using the escape key

src/Components/Popover/Base/PopoverDialog.tsx
```diff
@@ -146,6 +147,20 @@ export const PopoverDialog = forwardRef<PopoverDialogRef, PopoverDialogProps>(
...
+    // enableDefaultStyles prop changes
+    useEffect(() => {
+      handleUpdateStyles()
+    }, [enableDefaultStyles])
+
+    // Ensure dialog element is in focus on mount
+    // in order to enable escape key behavior
+    useEffect(() => {
+      if (dialogRef.current) {
+        dialogRef.current.focus()
```
mavarius, line 160: Could this inadvertently pull focus from an element within the dialog, like a button or input field?

alexpaxton: That's a good question... I'll try building an example of this in storybook

alexpaxton:
I updated `PopoverDialog` to only pull focus if no other element is in focus, which you can apparently check using `document.activeElement`. Also built a story that verifies popover focusing works with and without children using autofocus

PR #516, opened 2020-07-08
Introduce List component family

alexpaxton:
# Why?

Within the currently available components there is no generic "list" component family. The closest components are the `DropownMenu` and `DropdownItem`s but they are very opinionated and not designed for contexts outside of dropdowns.

In the InfluxDB UI there are a handful of components that resemble a list but each are slightly different. Between those cases, dropdowns, and right click menus there is an opportunity to consolidate and standardize. Why not have one family of components that can serve all these contexts and more?

### Changes

- Introduce `List` family of components

### Notes

- Instead of creating a separate version of `ListItem` intended for use with `<a>` or `<Link>` there is a prop for that. Due to how the children of `ListItem` and itself relate having an extra DOM element in between throws off the styling.
- Compared to the dropdown family the `List` components are much more composable
- Favoring `backgroundColor` and `gradient` as the means of colorization instead of `ComponentColor` enum
- Unlike dropdowns, each component can be colored individually
  - The styles are not dependent on DOM structure anymore
- Once this PR is merged, `List` can begin to replace parts of other components such as `RightClick`, `SelectDropdown`, and `MultiSelectDropdown`

### Screenshots

Screen Shot 2020-07-15 at 2 19 17 PM
Screen Shot 2020-07-15 at 2 19 28 PM
Screen Shot 2020-07-15 at 2 19 36 PM
Screen Shot 2020-07-15 at 2 19 55 PM
Screen Shot 2020-07-15 at 2 20 13 PM
Screen Shot 2020-07-15 at 2 20 26 PM

### Checklist
Check all that apply

- [x] Tests pass


alexpaxton:
Todo:

- [ ] Ensure styles exist for disabled state of items
- [ ] Make `onClick` optional for static display of information

alexpaxton:
Switched to using `InfluxColors` / `Gradients` as the coloration method for list items (instead of `ComponentColor`).

- [ ] Polish the selected state of indicators with gradients
- [ ] Rename some contexts and add comments in the code
- [ ] Revisit `ListLinkItem` approach

src/Components/List/ListEmpty.tsx
```diff
@@ -0,0 +1,54 @@
...
+export interface ListEmptyProps extends StandardFunctionProps {
+  /** Controls whether the text contents of this item wrap or not */
+  wrapText?: boolean
+  /** Size of this component */
+  size?: ComponentSize
+}
+
+export type ListEmptyRef = HTMLDivElement
+
+export const ListEmpty = forwardRef<ListEmptyRef, ListEmptyProps>(
```
mavarius, line 17: I wonder if we should call this `ListEmptyState` in keeping with then `EmptyState` component use.

src/Components/List/ListItem.tsx
```diff
@@ -0,0 +1,201 @@
...
+  /** Overrides backgroundColor, fills background with gradient */
+  gradient?: Gradients
+}
+
+export type ListItemRef = HTMLDivElement
+
+export const ListItemContext = createContext<ListItemContextProps>({
+  selected: false,
+  size: ComponentSize.Small,
+  listItemBackgroundColor: undefined,
```
mavarius, line 67: Is `undefined` necessary here? Seems redundant to define a default prop as "undefined".

alexpaxton: The context needs some initial value set. I could do `null` instead. I could be wrong but I think it will complain if I don't set a value

mavarius: Ya, defining null would be better than defining undefined.

src/Components/List/List.tsx
```diff
@@ -0,0 +1,183 @@
...
+export interface ListContextProps {
+  listBackgroundColor?: InfluxColors | string
+  listGradient?: Gradients
+  listContrastColor?: InfluxColors | string
+}
+
+export type ListItemRef = HTMLDivElement
+
+export const ListContext = createContext<ListContextProps>({
+  listBackgroundColor: undefined,
```
mavarius, line 36: we can switch these to `null` too.

PR #530, opened 2020-08-03
Refactor Portal-Based Components

alexpaxton:
### Problem

- Having a dedicated portal element for every instance of a portal-based component can result in a lot of re-rendering

### Changes

- Since the previous decision was to enable portal-based components to be used independently of the rest of Clockface, this changeset preserves that goal
- Introduce `usePortal` hook and export alongside components
  - Refactor portal-based components to use `usePortal` hook
  - `usePortal` creates a portal element as soon as it is needed, then it remains in place for all subsequent components to use.
  - Ensures that a single portal element is always visible
- Refactor `z-index` variables
  - Notifications will always be on top
  - Elements sent to the portal will stack in the order they were sent
    - This follows the users journey and feels natural
- Create storybook section called "Sandbox" for stories that test interplay between multiple types of components

### Bonus

- Since there is an ever-present portal element, components that rely on mount/unmount transitions animate much more smoothly

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/src/Components/Overlay/Overlay.scss
+++ b/src/Components/Overlay/Overlay.scss
@@ -5,10 +5,6 @@
    -----------------------------------------------------------------------------
 */
 
-.cf-overlay-portal {
-  @include portal-style($cf-z--overlays);
-}
-
 // Mixins for shared styles
 %overlay-styles {
   position: absolute !important;
@@ -67,13 +63,12 @@
   @include no-user-select();
 }
 
-
 .cf-overlay--title {
   font-family: $cf-overlay--title-font;
   font-size: $cf-overlay--title-size;
   font-weight: $cf-overlay--title-weight;
   color: $cf-overlay--title-color;
-  
+
   .cf-overlay--header__dismissable & {
     width: calc(100% - #{$cf-overlay--header-height - 50px});
   }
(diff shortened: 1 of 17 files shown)


src/Components/Overlay/Overlay.scss
```diff
@@ -5,10 +5,6 @@
    -----------------------------------------------------------------------------
 */
 
-.cf-overlay-portal {
-  @include portal-style($cf-z--overlays);
-}
-
 // Mixins for shared styles
 %overlay-styles {
   position: absolute !important;
```
Asalem1, line 10: *gasp*

src/Components/Overlay/Overlay.tsx
```diff
@@ -74,6 +48,15 @@ export const OverlayRoot: FunctionComponent<OverlayProps> = ({
...
   const overlay = (
     <>
+      <Transition
+        items={visible}
+        from={{opacity: 0}}
+        enter={{opacity: 0.7}}
+        leave={{opacity: 0}}
+        config={transitionConfig}
+      >
+        {visible => visible && (props => renderMaskElement(props))}
```
Asalem1, line 58: 🔥

src/Utils/index.ts
```diff
@@ -219,54 +219,6 @@ export const getScrollbarColorsFromTheme = (
...
-): void => {
-  const portalElement = document.getElementById(id)
-
-  if (portalElement) {
-    changeEventListener(portalElement)
-  }
-}
-
 export const getDictionary = (): string[] => {
   const hipsterIpsum =
```
Asalem1, line 223: this is my new favorite thing

PR #540, opened 2020-08-17
Fix usePortal Hook

### LINKED ISSUE #537: usePortal is broken
Issues:

- Instances of nested portal-based components render out of order
  - ex: in check editor overlay the always-visible popover renders underneath the overlay because it mounts before the overlay does
- The `addEventListenerToPortal` function is causing some typescript issues
- The `intersectionObserver` in `Popover` is also causing some issues

alexpaxton:
Closes #537

### Changes

- Rework how event listeners get added to portals
- Allow elements passed in to `usePortal` to be forced to the top
  - Only needed this for 1 instance in InfluxDB
  - Not very elegant but allows e2e test to pass
- Make popovers less likely to lose focus
- Guard against trigger not yet existing before applying intersection observer

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/src/Components/Popover/Base/Popover.tsx
+++ b/src/Components/Popover/Base/Popover.tsx
@@ -75,7 +75,7 @@ export const PopoverRoot = forwardRef<PopoverRef, PopoverProps>(
       contents,
       className,
       triggerRef,
-      forceToTop,
+      forceToTop = false,
       caretSize = 8,
       visible,
       disabled = false,
--- a/src/Components/Popover/Base/Popover.tsx
+++ b/src/Components/Popover/Base/Popover.tsx
@@ -55,6 +55,8 @@ export interface PopoverProps extends StandardFunctionProps {
   triggerRef: RefObject<any>
   /** Adds reasonable styles to popover dialog contents so you do not have to */
   enableDefaultStyles?: boolean
+  /** Ensures the popover appears above all other portal elements */
+  forceToTop?: boolean
 }
 
 export type PopoverRef = PopoverDialogRef
@@ -73,6 +75,7 @@ export const PopoverRoot = forwardRef<PopoverRef, PopoverProps>(
       contents,
       className,
       triggerRef,
+      forceToTop,
       caretSize = 8,
       visible,
       disabled = false,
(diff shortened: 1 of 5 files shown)


src/Components/Popover/Base/Popover.tsx
```diff
@@ -55,6 +55,8 @@ export interface PopoverProps extends StandardFunctionProps {
   triggerRef: RefObject<any>
   /** Adds reasonable styles to popover dialog contents so you do not have to */
   enableDefaultStyles?: boolean
+  /** Ensures the popover appears above all other portal elements */
+  forceToTop?: boolean
```
mavarius, line 59: We can hope this is used sparingly but I feel like once people start to use this in an application subsequent popovers will also end up having it applied. Seems like a slippery slope to all elements having this applied which would defeat the purpose. This isn't a problem for the library as much as `!important` is a problem for CSS, but maybe we should add some documentation around it to promote good practices.

src/Components/Popover/Base/PopoverDialog.tsx
```diff
@@ -156,11 +158,13 @@ export const PopoverDialog = forwardRef<PopoverDialogRef, PopoverDialogProps>(
     // Ensure dialog element is in focus on mount
     // in order to enable escape key behavior
     useEffect(() => {
-      const okayToPullFocus = !document.activeElement
+      const okayToPullFocus =
+        document.activeElement?.tagName !== 'INPUT' &&
```
mavarius, line 162: Nice, look at you on the cutting edge with that optional chaining.

src/Utils/portals.tsx
```diff
@@ -58,8 +59,17 @@ const getNotificationContainer = (
...
-  const addElementToPortal = (element: ReactNode): ReactPortal => {
-    return createPortal(element, portal)
+  const addElementToPortal = (
+    element: ReactNode,
+    forceToTop?: boolean
+  ): ReactPortal => {
+    let registeredElement = element
+
+    if (forceToTop) {
+      registeredElement = <div style={{zIndex: 9999}}>{element}</div>
```
mavarius, line 69: Wrapping it in a div conditionally might cause some unexpected behavior, especially if the element is used inline, or if a predictable order of elements is being used by some third-party applications like Google Analytics.

PR #551, opened 2020-10-14
Introduce TimeInput

alexpaxton:
A specialized input designed for specifying units of time. This is defined as a number + a unit `s, m, h, d, w, mo`

### Changes

- Introduce `TimeInput` as a composed variation on `Input`
- Create `TimeUnit` enum
- Modify `ButtonGroup` to also affect inputs placed inside
- Add `monospace` prop to `Input`

### Screenshots

Screen Shot 2020-10-14 at 10 58 11 AM

### Checklist
Check all that apply

- [x] Tests pass


mavarius (approved):
Looks good, though I wonder if we'll run into trouble hardcoding the units. We may end up needing a way to specify larger units like years or smaller units like milliseconds.

alexpaxton:
> Looks good, though I wonder if we'll run into trouble hardcoding the units. We may end up needing a way to specify larger units like years or smaller units like milliseconds.

@mavarius I had a similar thought, like maybe instead of an enum it could just be an array that can be overridden if you want more or less units. Not 100% sure how this is gonna be used yet, all I know is there will be inputs in the alerts UI that want units of time specifically

PR #554, opened 2020-10-28
fix(inputs): handle checkbox clicks/changes

OfTheDelmer:
### Changes

When an input checkbox is clicked it targets a div instead of the
actual input. This change uses a ref to the input to forward clicks on
the div to the input.

### Checklist
Check all that apply

- [x] Tests pass


mavarius:
Stale, no longer needed.

PR #563, opened 2021-01-08
fix: Makes rollup copy src specific to avoid src/dest conflict issue

### LINKED ISSUE #562: Source and Destination cannot be the same - yarn start
**Describe the bug**
Rollup watch fails because Image copying source and destination cannot be the same. 

**To Reproduce**
1. run `yarn start`
2. Observe error message in console: "Source and destination cannot be the same"

**Expected behavior**
- No error message! Rollup should successfully watch and build changes.

image

genehynson:
Closes #562

### Changes

Changes the `src` argument to the Rollup `copy` plugin to be more specific. The glob `*/**/Images/*` is too generic and can match the `dest` argument `dist/Images` causing error "Source and destination must not be the same". 

The only image I found in the repo was under `src` directory so I changed the the glob to `src/**/Images/*`. 

image

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -28,7 +28,7 @@ let plugins = [
     targets: [
       {src: 'src/Styles/Fonts', dest: 'dist'},
       {src: 'src/Styles/variables.scss', dest: 'dist'},
-      {src: '*/**/Images/*', dest: 'dist/Images'}
+      {src: 'src/**/Images/*', dest: 'dist/Images'}
     ],
   }),
   sourceMaps(),


drdelambre (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 #### 2.4.2 (Unreleased)
 
+- [#563](https://github.com/influxdata/clockface/pull/563): Fixes issue with Rollup copy plugin where src and destination could be the same, causing development issues.
```
hoorayimhelping, line 5: I don't think we need this in the changelog since it's a development issues. We usually only put user-facing changes in changelogs

hoorayimhelping (approved):
looks good with the changelog caveat

alexpaxton (approved):

PR #568, opened 2021-02-01
fix:  labels on toggle buttons are now clickable

### LINKED ISSUE #566: Toggle button labels do not toggle the button
when you have toggle buttons like:

Screen Shot 2021-01-29 at 4 17 26 PM

you expect that clicking on the "Vertical" label will activate the vertical toggle/radio button and turn off the Horizontal button.
that doesn't work.

it is because of custom onClick handlers in the toggle implementation  that this doesn't work; so switching the onClick handler to encompass the entire component to fix this.

worked on this with @mavarius 
example code for the screenshot above; note that VERTICAL and HORIZONTAL are string constants

    <FlexBox
      direction={FlexDirection.Column}
      margin={ComponentSize.Large}
      alignItems={AlignItems.FlexStart}
      style={{marginBottom: 12}}
     >
      <p>Legend Orientation</p>
         <Toggle
           tabIndex={1}
           value="horizontal"
           id="horizontal-legend-orientation"
           name="horizontal"
           checked={legendOrientation === HORIZONTAL}
           onChange={() => handleSetOrientation(HORIZONTAL)}
           type={InputToggleType.Radio}
           size={ComponentSize.ExtraSmall}
           color={ComponentColor['Primary']}
           appearance={Appearance['Outline']}
         >

        <InputLabel
          active={legendOrientation === 'horizontal'}
          htmlFor="horizontal-legend-orientation"
        >
          Horizontal
        </InputLabel>
      </Toggle>
      <Toggle
        tabIndex={2}
        value="vertical"
        id="vertical-legend-orientation"
        name="vertical"
        checked={legendOrientation === VERTICAL}
        onChange={() => handleSetOrientation(VERTICAL)}
        type={InputToggleType.Radio}
        size={ComponentSize.ExtraSmall}
        color={ComponentColor['Primary']}
        appearance={Appearance['Outline']}
      >
        <InputLabel
          active={legendOrientation === 'vertical'}
          htmlFor="vertical-legend-orientation"
        >
          Vertical
        </InputLabel>
      </Toggle>
    </FlexBox>`

jrenee42:
also adding pretty-quick
         - you can just call 'yarn pretty-quick' at the command line, and only the files you changed will get prettier run on them;
	 it is so much quicker and easier to use  (takes 5-10 seconds instead of 2-3 minutes)

Closes #566 

### Changes

   the on-click listener is now on the entire div; so that clicking on the children will toggle as well
    (b/c of the explicit onclick handler, the 'for' linking the labels to the input does not toggle the button at all)

### DIFF
--- a/src/Components/Button/Documentation/ButtonBase.md
+++ b/src/Components/Button/Documentation/ButtonBase.md
@@ -2,7 +2,11 @@
 
 The Base Button is intended for custom use cases. It renders children inside itself to facilitate many use cases.
 
+Note that if you are looking for radio buttons, look in Components->Inputs -> Multiple Choice Form.
+our Implementation of Radio Buttons are Toggle Buttons.
+
 ### Usage
+
 ```tsx
 import {ButtonBase} from '@influxdata/clockface'
 ```
--- a/src/Components/Inputs/Toggle.tsx
+++ b/src/Components/Inputs/Toggle.tsx
@@ -163,8 +163,17 @@ export const Toggle = forwardRef<ToggleRef, ToggleProps>(
 
     const title = disabled ? disabledTitleText : titleText
 
+    // putting onClick handler on the containing div,
+    // so that clicking on the children will toggle as well
+    // (b/c of the explicit onclick handler, the 'for' linking the labels
+    // to the input does not toggle the button at all)
     return (
-      <div className={toggleClass} style={style} ref={containerRef}>
+      <div
+        className={toggleClass}
+        style={style}
+        ref={containerRef}
+        onClick={handleClick}
+      >
         <input
           id={id}
           ref={ref}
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -102,5 +102,8 @@
     "fix": "eslint --fix 'src/**/*.{ts,tsx}'; prettier --config .prettierrc.json --check 'src/**/*.{ts,tsx}'",
     "storybook": "start-storybook -p 9001 -c .storybook",
     "publishStorybook": "storybook-to-ghpages --out=.out"
+  },
+  "dependencies": {
+    "pretty-quick": "^3.1.0"
```
TCL735, line 107: This a new dependency and the only one listed under "dependencies". Seems like it is in the wrong place. I think we should move this to devDependencies.

TCL735 (changes requested):

package.json
jrenee42: so instead of doing

`% yarn add pretty-quick`

do:

`% yarn add --dev pretty-quick`

correct? @TCL735

TCL735: yeah i think so

TCL735 (approved):

PR #570, opened 2021-02-03
feat: expose min height that allows container to be visible and onscroll

### LINKED ISSUE #523: Expose onScroll in all components that utilize DapperScrollbars
There are a handful of scrollable components that obscure some of the props from `DapperScrollbars` such as `onScroll`. For example `PageContents` does not enable any `onScroll` based events

ShmuelLotman:
Closes #523 

### Changes
This exposes a minHeight prop that allows us to actually see the container when we have it scrollable. Before it was completely hidden. This was a bug IMO, but i could be missing something. Putting it here to look it over, but as far as i could tell, allowing the user to input the minHeight both adds flexibility to the component and addresses an issue here. 
// Describe what you changed

### Screenshots

### Checklist
Check all that apply

- [] Updated documentation to reflect changes
- [ ] Tests pass

### DIFF
--- a/src/Components/Page/PageContents.tsx
+++ b/src/Components/Page/PageContents.tsx
@@ -3,7 +3,10 @@ import React, {forwardRef} from 'react'
 import classnames from 'classnames'
 
 // Components
-import {DapperScrollbars} from '../DapperScrollbars/DapperScrollbars'
+import {
+  DapperScrollbars,
+  FusionScrollHandler,
+} from '../DapperScrollbars/DapperScrollbars'
 
 // Types
 import {StandardFunctionProps, ComponentSize} from '../../Types'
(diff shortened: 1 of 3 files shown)


mavarius:
Can you add a screenshot to the PR

src/Components/Page/PageContents.tsx
```diff
@@ -17,6 +20,10 @@ export interface PageContentsProps extends StandardFunctionProps {
   autoHideScrollbar?: boolean
   /** Controls the gutters (left and right margins) */
   gutters?: ComponentSize
+  /** Exposes custom onScroll handler */
+  onScroll?: FusionScrollHandler
+  /** Allows user to customize visible height of container */
+  visibleHeight?: string
```
mavarius, line 26: Couldn't we just define `min-height: 100px` in the CSS?

ShmuelLotman: We could, but i want the user to be able to pass one.

TCL735:
Stale.

PR #573, opened 2021-02-08
Feat: dropdown enhancements

jrenee42:
supporting/required for 20195 in the ui repository

### Changes

 Adding 2 optional properties, if those properties are not there then the. dropdown works exactly as it did before these changes

from the documentation:

 onClickAway :  Optional method that is triggered when the user clicks outside of/away from the dropdown 

  menuOpen:  string
  
  optional way for the owner of this dropdown to set the menu state:
 'open':  the menu will be open
 'closed': the menu will be closed
 any other string will be ignored.

 and the menu state (open/closed) only changes when the string changes.
 so the owner is responsible for resetting the string
 (if the string is set to 'open', and then the user closes it, and the code sets it to open again,
 unless the owner/code sets it to something else in between (like 'ignore' or 'close'),
 then nothing will happen- the menu will not open)
 

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/src/Components/Dropdowns/Dropdown.tsx
+++ b/src/Components/Dropdowns/Dropdown.tsx
@@ -30,18 +30,18 @@ export interface DropdownProps extends StandardFunctionProps {
   /** Optional method that is triggered when the user clicks outside of/away from the dropdown */
   onClickAway?: () => void
   /**
-   * optional way for the owner of this dropdown to set the menu status:
+   * optional way for the owner of this dropdown to set the menu to be open or closed:
    * 'open':  the menu will be open
    * 'closed': the menu will be closed
    * any other string will be ignored.
    *
-   * and the menu state (open/closed) only changes when this value changes.
+   * and the menu  (open/closed) only changes programmatically when this value changes.
    * so the owner is responsible for resetting this value
    * (if the string is set to 'open', and then the user closes it, and the code sets it to open again,
    * unless the code sets it to something else in between (like null or 'close'),
    * then nothing will happen- the menu will not open)
    * */
-  setMenuStatus?: MenuStatus
+  menuOpen?: MenuStatus
 }
 
 export type DropdownRef = HTMLDivElement
@@ -57,7 +57,7 @@ export const DropdownRoot = forwardRef<DropdownRef, DropdownProps>(
       dropUp = false,
       className,
       onClickAway,
-      setMenuStatus,
+      menuOpen,
     },
     ref
   ) => {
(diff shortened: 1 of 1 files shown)


src/Components/Dropdowns/Dropdown.tsx
```diff
@@ -22,6 +22,21 @@ export interface DropdownProps extends StandardFunctionProps {
...
+   * 'closed': the menu will be closed
+   * any other string will be ignored.
+   *
+   * and the menu state (open/closed) only changes when the string changes.
+   * so the owner is responsible for resetting the string
+   * (if the string is set to 'open', and then the user closes it, and the code sets it to open again,
+   * unless the code sets it to something else in between (like 'ignore' or 'close'),
+   * then nothing will happen- the menu will not open)
+   * */
+  menuOpen?: string
```
hoorayimhelping, line 39: I expected `menuOpen` to be a boolean. I think `menuStatus` might be more descriptive. I think we should clamp the values with a TypeScript [string enum](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) since it's a little bit tighter way of doing what you're going for.

something like

```ts
enum MenuStatus {
  Open = 'open',
  Closed = 'closed'
}

menuStatus?: MenuStatus
```

src/Components/Dropdowns/Dropdown.tsx
```diff
@@ -48,8 +65,20 @@ export const DropdownRoot = forwardRef<DropdownRef, DropdownProps>(
 
     const handleCollapseMenu = (): void => {
       setExpandedState(false)
+      if (onClickAway) {
+        onClickAway()
+      }
     }
 
+    useEffect(() => {
+      if (menuOpen === 'closed') {
```
hoorayimhelping, line 74: If you go with the above enum, this should change to something like:

```ts
  if (menuOpen === MenuStatus.Closed) {
```

src/Components/Dropdowns/Dropdown.tsx
```diff
@@ -22,6 +22,21 @@ export interface DropdownProps extends StandardFunctionProps {
...
+   * 'open':  the menu will be open
+   * 'closed': the menu will be closed
+   * any other string will be ignored.
+   *
+   * and the menu state (open/closed) only changes when the string changes.
+   * so the owner is responsible for resetting the string
+   * (if the string is set to 'open', and then the user closes it, and the code sets it to open again,
+   * unless the code sets it to something else in between (like 'ignore' or 'close'),
+   * then nothing will happen- the menu will not open)
+   * */
```
hoorayimhelping, line 38: @mavarius just wanted to get your eyes on this in case you have a strong opinion about flowerbox comments. I don't have problems with them, but I tend to ignore them, because they tend to get out of date, which is usually people's reservation with them...

hoorayimhelping (commented):
some stylistic comments, but I'll let Daniel or Slotman approve or request changes

src/Components/Dropdowns/Dropdown.tsx
jrenee42: then we would need a third one, for 'ignore'.  because it's not just the owner setting the open/close status, the user is controlling it too.

jrenee42: ...or could use null instead...  'setMenuStatus'  is a better name

hoorayimhelping (approved):
we talked offline about a small variable name change, other than LGTM!

PR #581, opened 2021-03-02
chore: improve lighthouse accessibility score

genehynson:
Related: https://github.com/influxdata/ui/issues/520

### Changes

Made some small tweaks to improve the accessibility and best practice scores from lightouse. 

1) Changed EmptyState's header from an h4 to an h2 so that our headers decrease sequentially. 
1) Changed EmptyState's font color to improve contrast ratio
1) Removed the tabIndex=1 from portals.

Also see this UI PR: https://github.com/influxdata/ui/pull/764

### Checklist

- [x] Tests pass

### DIFF
--- a/src/Utils/portals.tsx
+++ b/src/Utils/portals.tsx
@@ -8,7 +8,6 @@ const createPortalElement = (): HTMLElement => {
   const portalElement = document.createElement('div')
   portalElement.setAttribute('class', portalElementID)
   portalElement.setAttribute('id', portalElementID)
-  portalElement.setAttribute('tabindex', 'null')
 
   document.body.appendChild(portalElement)
 
--- a/src/Utils/portals.tsx
+++ b/src/Utils/portals.tsx
@@ -8,7 +8,7 @@ const createPortalElement = (): HTMLElement => {
   const portalElement = document.createElement('div')
   portalElement.setAttribute('class', portalElementID)
   portalElement.setAttribute('id', portalElementID)
-  portalElement.setAttribute('tabindex', '1')
+  portalElement.setAttribute('tabindex', 'null')
 
   document.body.appendChild(portalElement)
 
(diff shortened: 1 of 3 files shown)


src/Utils/portals.tsx
```diff
@@ -8,7 +8,7 @@ const createPortalElement = (): HTMLElement => {
   const portalElement = document.createElement('div')
   portalElement.setAttribute('class', portalElementID)
   portalElement.setAttribute('id', portalElementID)
-  portalElement.setAttribute('tabindex', '1')
+  portalElement.setAttribute('tabindex', 'null')
```
genehynson, line 11: I'm not sure what Alex's intention was for adding this tabIndex in this PR https://github.com/influxdata/clockface/pull/540 🤔

hoorayimhelping: do you know what this tabindex does and/or what removing it does?

genehynson: My understanding is that saying `tabIndex=1` means that it will be put after all the elements with `tabIndex=0` (which is the default). But lighthouse complains that you shouldn't force the ordering of tabbing through elements because it could become confusing if the tabbing is unnatural (e.g. starting with an element at the bottom of the page).

I'm not sure if `null` is the correct solution here but it does keep lighthouse from complaining. Another option would just be to remove this line. I just don't know what the intention was with setting it to 1 in the first place 🤔

mavarius: I think this is because the portals are not following the natural order of the page. This is what popovers and modals are displayed in.

genehynson: Ok so I tested it out (didn't originally realize portals were for modals/popups) and the difference is with `tabIndex=1` the modal is _last_ in the "tabbing" order. Without this line, the modal is _first_. To me it makes sense for the modal to be first because it's the component currently in focus on the screen

hoorayimhelping: makes sense to me

hoorayimhelping (approved):
great work and follow-up!

Issue #591, opened 2021-03-15
FormValidationElement should have option for pre-validating inputs


TCL735:
**Is your feature request related to a problem? Please describe.**
As a user, the UI might have a page that identifies invalid resources that require me to take an action, such as renaming something. When I click the resource to rename it, the input field should start in error state. In other words, the input should be validated on first render or pre-validated.

**Describe the solution you'd like**
FormValidationElement should include a pre-validation mode. Or possibly Form can include an entirely new element that does pre-validation

**Describe alternatives you've considered**
The only workaround at the moment is having `FormValidationElement`'s child function do a one-time validation upon first render. This seems like additional boilerplate that needs to be repeated in every use case that calls for pre-validation.

PR #592, opened 2021-03-15
adding a prevalidation mode for usecases when you want validation on …

daviddkkim:
…initial render

Closes #

https://github.com/influxdata/clockface/issues/591

### Changes

// Describe what you changed
As it is now, validation does not happen until a user interacts with the input.

Added a pre-validation mode (boolean prop) to allow validations to happen on initial render. 

Example usecase:
 - If a user is prompted to change a bad variable, we want to load the input form with an error state rather to prompt them to fix it. 
 
 @TCL735 has a great explanation for the use case if needed more clarity. 

### Screenshots

Screen Shot 2021-03-15 at 6 37 04 PM

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### 2.6.4 (Unreleased)
+
+- [#592](https://github.com/influxdata/clockface/pull/592): Add prevalidation mode for formValidationElement
+
+
 ### 2.6.3 (2021-3-15)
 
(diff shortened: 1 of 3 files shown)


TCL735:
Nice! :fire:

i think this works for my use case.

TCL735 (approved):

mavarius (approved):
Good to go

PR #594, opened 2021-03-16
chore: updated FormElement props and FormElementError to allow for custom testIDs to be passed in for the formelementerror

Asalem1:
### Changes

Updated the FormElement to allow for an additional testID to be passed to the FormElement that will be propagated onto the FormErrorMessage. This will allow for more customized testing on error messages. 

### Screenshots

### Checklist
Check all that apply

- [ ] Tests pass

### DIFF
--- a/src/Components/Form/FormElement.tsx
+++ b/src/Components/Form/FormElement.tsx
@@ -23,6 +23,8 @@ export interface FormElementProps extends StandardFunctionProps {
   required?: boolean
   /** Useful for associating a label with an input */
   htmlFor?: string
+  /** ID for Error Message for Integration Tests */
+  errorMessageTestId?: string
 }
 
 export type FormElementRef = HTMLLabelElement & HTMLDivElement
@@ -41,6 +43,7 @@ export const FormElement = forwardRef<FormElementRef, FormElementProps>(
       labelAddOn,
       errorMessage,
       testID = 'form--element',
+      errorMessageTestId,
     },
     ref
   ) => {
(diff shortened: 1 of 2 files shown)


src/Components/Form/FormElement.tsx
```diff
@@ -57,7 +60,10 @@ export const FormElement = forwardRef<FormElementRef, FormElementProps>(
         )}
         {children}
         {!!errorMessage && (
-          <FormElementError testID={testID} message={errorMessage} />
+          <FormElementError
+            errorMessageTestId={errorMessageTestId}
```
TCL735, line 64: `data-testid={errorMessageTestId}`

TCL735 (changes requested):

src/Components/Form/FormElement.tsx
Asalem1: Aw, I see what you're saying

TCL735: Oops... my mistake. `<FormElementError>` is not atomic and already has a prop for taking data-testid which is actually called "testID".

Can we please change this to: `testID={errorMessageTestId}` and then revert the changes in the other file.

src/Components/Form/FormElementError.tsx
```diff
@@ -22,7 +23,7 @@ export const FormElementError = forwardRef<
       style,
       className,
       message = '\u00a0\u00a0',
-      testID = 'form--element-error',
+      errorMessageTestId = 'form--element-error',
```
TCL735: My mistake. All changes in this file should be reverted. No changes should be required here.

TCL735 (changes requested):

TCL735 (approved):

PR #613, opened 2021-04-25
Adding support for help text -> error text conversion

### LINKED ISSUE #232: Validation errors should not shift input fields
**Describe the bug**
When error messages appear below input fields, everything below them shifts.

**Expected behavior**
The layout of the form should remain static when validation errors are present.

daviddkkim:
Closes #232

### Changes

// Describe what you changed
I added new styling container on top of the form error message so that when an error message appears, it does not take up spacing.

### Screenshots

https://user-images.githubusercontent.com/12561526/116012070-24e0cd80-a5dd-11eb-83b4-38986f2f2930.mov

There is one caveat. This means that developers using clockface have to add in the right spacing after the input element so that this does not happen:

Screen Shot 2021-04-25 at 3 33 44 PM

Or I can make it so that it has built in spacing out of the box.  @mavarius do you have thoughts on which one might be preferable? 

I see good reasons for having pre-defined spacing to prevent the overlap (easy out of the box) and also not having it (more freedom). 

### Checklist
Check all that apply

- [ ] Tests pass

### DIFF
--- a/src/Components/Form/Documentation/Form.stories.tsx
+++ b/src/Components/Form/Documentation/Form.stories.tsx
@@ -234,6 +234,7 @@ formStories.add(
         <Form.ElementError
           ref={formElementErrorRef}
           message={text('errorMessage', 'Error Message')}
+          withHelper={boolean('withHelper', false)}
         />
         <div className="story--test-buttons">
           <button onClick={logRef}>Log Ref</button>
--- a/src/Components/Form/Form.scss
+++ b/src/Components/Form/Form.scss
@@ -36,6 +36,12 @@
   @extend %no-user-select;
 }
 
+.cf-form--element-error-container {
+  position: relative;
+  height: 0px;
+  width: 100%;
+}
+
 .cf-form--label {
   width: 100%;
   color: $cf-label--default;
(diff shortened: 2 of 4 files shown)


mavarius:
I think having the built-in spacing is good. It means all our forms will be a bit more vertically spaced, but that might be ok. Just review how it looks across the application to avoid any unforeseen side effects.

daviddkkim:
Example1Spacing
LabelSpacing


@mavarius  So here's a snippet of what the forms will look like with "built-in" spacing. I'm starting to think that we should leave the error message spacing behavior alone because built-in spacing throws off the vertical rhythm of the form elements and without built-in spacing, the error message overlaps with the component below.  

But I am all for changing the behavior of when there is help text + error text! 

What do you think about the spacing issue?

mavarius:
We could reduce the distance between the bottom of the error text and the top of the next element. If we absolute position the error text, then we can maintain a consistent distance between elements and have just a little less space when the error text is present without overlapping text.

Imagine this, but with 4px between the bottom of the error text and the top of the next label.
image

daviddkkim:
Yeah I like this. More spacing less crowded. 
Frame 1

Issue #616, opened 2021-04-28
Add optional text easy clearing in inputs


daviddkkim:
**Is your feature request related to a problem? Please describe.**

There's been a request from internal users to be able to clear text with a click of an icon(X). 

**Describe the solution you'd like**

~- When there is no text in the input, the X icon does not exist.~
~- When there is text to be cleared, the X icon appears.~ (Daniel brought up a great point that this is not accessibility friendly)
- When the input is focused, the X icon appears.
- When the X icon is focused, the X icon stays appeared.
- When neither the X icon or the input is focused, the X icon does not exist.

- The X Icon follows the same color scheme, hover, and active state patterns ( Color changes, transitions .. etc. Most, if not all components, have this behavior so please reference them) 

- The X icon is keyboard navigable (Tabs) to align with Accessibility standards.

**Describe alternatives you've considered**

**Additional context**

This is from this issue
-[] influxdata/ui#898

jrenee42:
>>When neither the X icon or the input is focused, the X icon does not exist.

then how will the users discover this feature?  what if they enter text, click away, then the 'x' disappears, then they want to clear it, so then they click in the field, and then it appears?  that is non-intuitive.

daviddkkim:
> > > When neither the X icon or the input is focused, the X icon does not exist.
> 
> then how will the users discover this feature? what if they enter text, click away, then the 'x' disappears, then they want to clear it, so then they click in the field, and then it appears? that is non-intuitive.

Cross-posting here and the PR


Good point. I think there is a middle ground here between my first suggestion of making the icon appear on text and Daniel's suggestion of making icon appear on focus for accessibility reasons.

How about

When the input is focused, the X icon appears. When the X icon is focused, the X icon stays appeared. When neither X icon or the input is focused but the text value is non empty, the X icon stays appeared. If neither X icon or the input is focused AND the text value is empty, the X icon disappears.
@mavarius @jrenee42 thoughts?

kristinarobinson:
Closed UI#898, since issue is solely a clockface issue.

mavarius:
done

PR #617, opened 2021-05-03
Jill wip/clearable button

jrenee42:
for ui #898
https://app.zenhub.com/workspaces/applicationui-5e3b05772922e316b3a210a4/issues/influxdata/ui/898

### Changes

adds an 'x' that shows up in the right of the textfield if and only if:
1) there is content in the textfield
2) the 'onClear' property is present; with a function to be activated when the button is clicked

the button is reachable by tabbing, so the user can type in data, then click 'tab' and the press 'return' to clear the field without ever having to use the mouse.

it uses the dismiss button, with some styling changes to get it to fit nicely in the field.

if the field is large, then the button is sized up, else it stays in its default, smaller size.

Added to storybook as well; so added to the documentation.

### Screenshots

https://user-images.githubusercontent.com/3900960/116930539-e1660f00-ac2d-11eb-8841-0810f700ceb4.mov

### Checklist
Check all that apply

- [ ] Tests pass


daviddkkim (changes requested):
https://github.com/influxdata/clockface/issues/616

> When the input is focused, the X icon appears. When the X icon is focused, the X icon stays appeared. When neither the X icon or the input is focused, the X icon does not exist.

src/Components/Inputs/Input.tsx
```diff
@@ -173,6 +182,19 @@ export const Input = forwardRef<InputRef, InputProps>(
 
     const iconElement = icon && <Icon glyph={icon} className="cf-input-icon" />
 
+    const clearClasses = classnames('cf-input-clear-btn', {
+      large: size === ComponentSize.Large,
+    })
+
+    const clearElement = correctlyTypedValue && onClear && (
+      <DismissButton
+        style={clearBtnStyle}
```
daviddkkim, line 191: Do we need to expose this style prop? I think we could just use color = {ComponentColor.default}. I don't see where we would need to change the color or style of the dismiss button within the input but if you can think of concrete future usecases for it, would love to hear it

src/Components/Inputs/Input.scss
```diff
@@ -82,14 +82,15 @@
...
-  z-index: $cf-input--status-z;
-  transition: color 0.25s ease;
   font-size: 1.1em;
+  transform: translate(-50%, -50%);
 }
 
-.cf-input-icon {
-  transform: translate(-50%, -50%);
+.cf-input-icon,
+.cf-input-clear-btn {
```
daviddkkim, line 90: This might be a .cf-input__clear-btn based on BEM methodology but I could be wrong

src/Components/Inputs/Input.scss
```diff
@@ -101,6 +102,38 @@
...
+  }
+
+  &.cf-dismiss-button {
+    transform: none;
+    top: 4px;
+    right: 4px;
+  }
+
+  &.large {
+    width: 32px;
```
daviddkkim, line 123: Let's use variables instead of hard coding the width and heights if we can.

src/Components/Inputs/Input.scss
```diff
@@ -101,6 +102,38 @@
...
+    opacity: 0;
+  }
+
+  &.cf-dismiss-button {
+    transform: none;
+    top: 4px;
+    right: 4px;
+  }
+
+  &.large {
```
daviddkkim, line 122: I believe the possible sizes of inputs are :
- extra small,
- small,
-medium
- large , 

I think you've only accounted for small and large. Try a medium or extra small input on your storybook, I imagine the X won't scale for those.

src/Components/Inputs/Input.tsx
daviddkkim: As for styling, I think the default color set is the subtlety we are looking for rather than the primary color that comes out of box with the dismiss button component.

src/Components/Inputs/Input.tsx
```diff
@@ -173,6 +182,19 @@ export const Input = forwardRef<InputRef, InputProps>(
 
     const iconElement = icon && <Icon glyph={icon} className="cf-input-icon" />
 
+    const clearClasses = classnames('cf-input-clear-btn', {
+      large: size === ComponentSize.Large,
```
daviddkkim, line 186: need to account for extra small, and medium as well

src/Components/Inputs/Input.tsx
```diff
@@ -79,6 +80,13 @@ export interface InputProps extends StandardFunctionProps {
   inputStyle?: CSSProperties
   /** For use within a form, marks the input as required */
   required?: boolean
+  /**
+   * Function to be called when the input field is cleared;
+   * if not included then the 'x' clear button will NOT be added
+   * */
+  onClear?: () => void
+  /** optional extra styling for the clear button */
+  clearBtnStyle?: CSSProperties
```
daviddkkim, line 89: Bit of a redundant feedback here but Not sure if we need to expose this. Im trying to think of a case where we would need a special css treatment of the dismiss button and can't think of it.

src/Components/Inputs/Input.scss
```diff
@@ -101,6 +102,38 @@
...
+    width: 32px;
+    height: 32px;
+    right: 9px;
+    top: 6px;
+
+    & .cf-dismiss-button--x {
+      &:before,
+      &:after {
+        width: 20px;
```
daviddkkim, line 131: Let's use variables instead of hard coding the width and heights if we can.

daviddkkim (changes requested):

src/Components/Inputs/Documentation/Inputs.stories.tsx
```diff
@@ -152,6 +152,49 @@ inputsBaseStories.add(
...
 )
+inputsBaseStories.add(
+  'Input (Clearable Text)',
+  () => {
+    const [value, setValue] = useState<string>('hello world')
+
+    return (
+      <div className="story--example">
+        <Input
+          min={number('min', 0)}
```
mavarius, line 163: Did you mean `minLength` & `maxLength`? `min` & `max` are for `type="number"`.

src/Components/Inputs/Input.tsx
```diff
@@ -12,18 +12,19 @@ import classnames from 'classnames'
 // Components
 import {Icon} from '../Icon/Base/Icon'
 import {StatusIndicator} from './StatusIndicator'
-
+import {DismissButton} from '../Button/Composed/DismissButton'
+//import {Button} from '../Button/Composed/Button'
```
mavarius, line 16: Remove if no longer used.

mavarius (commented):
Having the button disappear when you click it is a jarring experience. As long as the field or the button has focus then it should be visible.

image
You may need to prevent this from appearing when `type="number"`.


image
image
Ensure that the button is vertically centered regardless of the input size. It may help here to use the extra small and medium button sizes along with those input sizes.


image
The text should stop before the clear button otherwise it becomes obscured. See the `VisibilityInput` component for how this should be handled.

src/Components/Inputs/Input.tsx
jrenee42: are you saying that the color should. be changed from what it currently is?

src/Components/Inputs/Input.scss
jrenee42: i only did small (default) and large since those are the only use cases we need for the ui; I can add the others but we aren't using them at this point

src/Components/Inputs/Input.tsx
jrenee42: i needed it previously, then after adjusting for component size don't need it; but it was already there so I left it in.

src/Components/Inputs/Documentation/Inputs.stories.tsx
jrenee42: the 'min' and such I just copied from another storybook story.
(for the default, plain text input).

in terms of the button disappearing, @daviddkkim  specifically requested that the 'x' button only show when there is content.  so after pressing it, there is no content so it disappears.  easy to make it always appear tho.

jrenee42:
> #616
> 
> > When the input is focused, the X icon appears. When the X icon is focused, the X icon stays appeared. When neither the X icon or the input is focused, the X icon does not exist.

then how will the users discover this feature? what if they enter text, click away, then the 'x' disappears, then they want to clear it, so then they click in the field, and then it appears? that is non-intuitive.

src/Components/Inputs/Input.tsx
daviddkkim: Two things:
1.  I think the Default Color Dismiss button might be more suitable than the Primary Color Dismiss Button. 
2. I'm trying to understand the rationale for style={clearBtnStyle}. It seems like its just passing in undefined.
 In what scenarios would we want to pass in custom style to this dismiss button?

daviddkkim:
> > #616
> > > When the input is focused, the X icon appears. When the X icon is focused, the X icon stays appeared. When neither the X icon or the input is focused, the X icon does not exist.
> 
> then how will the users discover this feature? what if they enter text, click away, then the 'x' disappears, then they want to clear it, so then they click in the field, and then it appears? that is non-intuitive.

Good point. I think there is a middle ground here between my first suggestion of making the icon appear on text and Daniel's suggestion of making icon appear on focus for accessibility reasons.

How about 
- When the input is focused, the X icon appears. When the X icon is focused, the X icon stays appeared. When neither X icon or the input is focused but the text value is non empty, the X icon stays appeared. If neither X icon or the input is focused AND the text value is empty, the X icon disappears. 

@mavarius @jrenee42  thoughts?

src/Components/Inputs/Input.tsx
jrenee42: done.

src/Components/Inputs/Input.tsx
jrenee42: ...just removed it

jrenee42:
having it appear sometimes/not others may make styling tricky.

right now, the button always appears if the button is 'enabled' for the field.



> > > #616
> > > > When the input is focused, the X icon appears. When the X icon is focused, the X icon stays appeared. When neither the X icon or the input is focused, the X icon does not exist.
> > 
> > 
> > then how will the users discover this feature? what if they enter text, click away, then the 'x' disappears, then they want to clear it, so then they click in the field, and then it appears? that is non-intuitive.
> 
> Good point. I think there is a middle ground here between my first suggestion of making the icon appear on text and Daniel's suggestion of making icon appear on focus for accessibility reasons.
> 
> How about
> 
> * When the input is focused, the X icon appears. When the X icon is focused, the X icon stays appeared. When neither X icon or the input is focused but the text value is non empty, the X icon stays appeared. If neither X icon or the input is focused AND the text value is empty, the X icon disappears.
> 
> @mavarius @jrenee42 thoughts?

jrenee42:
> having it appear sometimes/not others may make styling tricky.
> 
> right now, the button always appears if the button is 'enabled' for the field.
> 
> > > > #616
> > > > > When the input is focused, the X icon appears. When the X icon is focused, the X icon stays appeared. When neither the X icon or the input is focused, the X icon does not exist.
> > > 
> > > 
> > > then how will the users discover this feature? what if they enter text, click away, then the 'x' disappears, then they want to clear it, so then they click in the field, and then it appears? that is non-intuitive.
> > 
> > 
> > Good point. I think there is a middle ground here between my first suggestion of making the icon appear on text and Daniel's suggestion of making icon appear on focus for accessibility reasons.
> > How about
> > 
> > * When the input is focused, the X icon appears. When the X icon is focused, the X icon stays appeared. When neither X icon or the input is focused but the text value is non empty, the X icon stays appeared. If neither X icon or the input is focused AND the text value is empty, the X icon disappears.
> > 
> > @mavarius @jrenee42 thoughts?

it actually looks ok to have the button hide when there is no content.  
(the code and styling part is trivial).

daviddkkim:
This looks almost ready @jrenee42. 

The styling is implemented slightly different than our input components with an icon. See 
Screen Shot 2021-07-13 at 4 30 42 PM

Screen Shot 2021-07-13 at 4 30 59 PM

This isn't a huge issue but I'd want to keep consistency.

Screen Shot 2021-07-13 at 4 34 11 PM

I'd want to see what other input's icon sizes are for each respective input size. I think for medium, we see 14 or 16 px. This one is 13.44px.  It looks like the width of the strokes in the X is determined by the height, if that's the case I think 1px or 2px would work best. We use 2px border radius throughout our styles i believe.

For hover state, active/focused state of the style, 
Please take a look at the [visibility input](http://localhost:9001/?path=/story/components-inputs-composed--visibility-input). I think the default state is fine as it is (except, instead of using #ffffff and opacity, I'd settle on an actual color in our design token) but the hover state and focus/active states of the dismiss button goes against how we style.  

I'm happy to make these finishing touches/style changes but I am occupied with million other tasks at the moment and won't be able to get around to it until I get some breathing room!

jrenee42:
> This looks almost ready @jrenee42.
> 
> The styling is implemented slightly different than our input components with an icon. See
> Screen Shot 2021-07-13 at 4 30 42 PM
> 
> Screen Shot 2021-07-13 at 4 30 59 PM
> 
> This isn't a huge issue but I'd want to keep consistency.
> 
> Screen Shot 2021-07-13 at 4 34 11 PM
> 
> I'd want to see what other input's icon sizes are for each respective input size. I think for medium, we see 14 or 16 px. This one is 13.44px. It looks like the width of the strokes in the X is determined by the height, if that's the case I think 1px or 2px would work best. We use 2px border radius throughout our styles i believe.
> 
> For hover state, active/focused state of the style,
> Please take a look at the [visibility input](http://localhost:9001/?path=/story/components-inputs-composed--visibility-input). I think the default state is fine as it is (except, instead of using #ffffff and opacity, I'd settle on an actual color in our design token) but the hover state and focus/active states of the dismiss button goes against how we style.
> 
> Screen Shot 2021-07-13 at 4 39 21 PM
> 
> I'm happy to make these finishing touches/style changes but I am occupied with million other tasks at the moment and won't be able to get around to it until I get some breathing room!

ok.  my concerns are that the dismiss button is a part of clockface, so maybe *that* needs changing too for consistency,
and having it be a button (other than an icon) means that it can accept tabs and focus for accessibility.

I will talk to Kristina and find out the priority on this (vs. other bugs/issues).  thanks for the review!

daviddkkim:
> > This looks almost ready @jrenee42.
> > The styling is implemented slightly different than our input components with an icon. See
> > Screen Shot 2021-07-13 at 4 30 42 PM
> > Screen Shot 2021-07-13 at 4 30 59 PM
> > This isn't a huge issue but I'd want to keep consistency.
> > Screen Shot 2021-07-13 at 4 34 11 PM
> > I'd want to see what other input's icon sizes are for each respective input size. I think for medium, we see 14 or 16 px. This one is 13.44px. It looks like the width of the strokes in the X is determined by the height, if that's the case I think 1px or 2px would work best. We use 2px border radius throughout our styles i believe.
> > For hover state, active/focused state of the style,
> > Please take a look at the [visibility input](http://localhost:9001/?path=/story/components-inputs-composed--visibility-input). I think the default state is fine as it is (except, instead of using #ffffff and opacity, I'd settle on an actual color in our design token) but the hover state and focus/active states of the dismiss button goes against how we style.
> > Screen Shot 2021-07-13 at 4 39 21 PM
> > I'm happy to make these finishing touches/style changes but I am occupied with million other tasks at the moment and won't be able to get around to it until I get some breathing room!
> 
> ok. my concerns are that the dismiss button is a part of clockface, so maybe _that_ needs changing too for consistency,
> and having it be a button (other than an icon) means that it can accept tabs and focus for accessibility.
> 
> I will talk to Kristina and find out the priority on this (vs. other bugs/issues). thanks for the review!

@jrenee42 
I think the color issue can be solved by just adding 
        color={ComponentColor.Tertiary}
 to the dismiss button.
 
 As for the dismiss button, yeah the dismiss button wasn't designed to be used inside of an input. That's why you won't find it usable for the input out of the box. The hardcoded height,width you have seems to not align with the icon sizes we use for inputs. 
 
 Thing to change:
 - Color to tertiary for dismiss button for a more muted look.
 - Icon sizes for input, extrasmall = 12x12, medium = 16 x 16 , large = 20 x 20.
 - all the hard coded values, try to find the appropriate design tokens in the variables scss file.

daviddkkim:
I can make this change or you can. I think I might have some time later in the week to tackle some of these but since this is not a priority for me, my attention on it will be sparse for full transparency.

PR #630, opened 2021-05-24
Storybook update to use main.js and add support for new addons

daviddkkim:
Closes #

### Changes
Current implementation of storybook uses config.js and addon.js which date back to Pre- version 5. This convention was thrown out with storybook 5.3 and they now encourage users to use main.js to configure storybook. We need to get onto main.js because config.js is no longer supported. 

I've also added new addons that will help us with accessibility and responsiveness.
Screen Shot 2021-05-23 at 9 37 45 PM
Screen Shot 2021-05-23 at 9 37 59 PM
localhost_9001__path=_story_atomic-alert--example

// Describe what you changed

### Screenshots

### Checklist
Check all that apply

- [ ] Tests pass


.storybook/main.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+    stories: ['../src/**/*.stories.tsx'],
+    addons: ['@storybook/addon-knobs',
+    'storybook-readme',
+    '@storybook/addon-a11y',
+    '@storybook/addon-viewport'
+],
+  };
```
mavarius, line 8: Did you run the linter? It'd likely automatically add the empty lines to the end of these files.

mavarius (approved):
Awesome, love this!

.storybook/main.js
daviddkkim: Ah I think I forgot to run linter after the last change. Running it now !

PR #633, opened 2021-05-27
add spacing to input field so that last few chars don't get hidden

### LINKED ISSUE #631: Email address text box on login cuts off last chars of email if there is an error
image (69)

If you enter an email / password combo that isn't valid, you get this message. The issue is that there is no way to see the last part of the email address which could contain a space or extra character when copy/pasting.

daviddkkim:
Closes #631 

### Changes

With Inputs, when status is not default, we add padding to the right equal to the size of the status indicator (the error symbol, valid symbol .. etc). This way no part of the last few characters will be hidden. I also reduced the range of the shadow effect of the status indicator so that the shadow doesn't cover the characters either.

// Describe what you changed

### Screenshots
inputFix

### Checklist
Check all that apply

- [ ] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 2.9.0 (2021-6-01)
 - [#632](https://github.com/influxdata/clockface/pull/632): Add disabledText field to Button-based components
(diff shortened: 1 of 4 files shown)


russorat (approved):
LGTM

mavarius (approved):
Make it so.

PR #646, opened 2021-07-20
fix: changed to correct classname

mtian725:
Closes https://github.com/influxdata/ui/issues/2086

ClassName for IndexList header for the arrow indicating sort direction was incorrect. Was originally `icon caret-down`, but is supposed to `cf-icon caret-down`. See `IndexList.scss` line 181.

## Before
Screen Shot 2021-07-20 at 3 55 54 PM

## After 
Screen Shot 2021-07-20 at 6 43 53 PM

### DIFF
--- a/src/Components/IndexList/IndexListHeaderCell.tsx
+++ b/src/Components/IndexList/IndexListHeaderCell.tsx
@@ -85,7 +85,7 @@ const SortIndicator: FunctionComponent<{sortable: boolean}> = ({sortable}) => {
   if (sortable) {
     return (
       <span className="cf-index-list--sort-arrow">
-        <span className="icon caret-down" />
+        <span className="cf-icon caret-down" />
       </span>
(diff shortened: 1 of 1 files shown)


daviddkkim:
@mtian725 good find. Do you have image of before?

hoorayimhelping (approved):
looks good to me.

Nitpick: I think your commit message can be a little more descriptive and contextual. `changed to correct classname` makes a lot of sense to me looking at it in a PR, but if i were to do something like `git log` and saw this amidst a bunch of other commits, I would probably have to examine the changes to understand what code changed. You can help me or any other person sidestep that work  by telling me what this change does in the commit message.

For example:
>fix: changed to correct classname

vs
>fix: updated classname on IndexList HeaderCell to show shorting icon

daviddkkim (approved):
LGTM

PR #657, opened 2021-08-12
chore: raising z-index of overlays so they are on top of everything

jrenee42:
changed z-index of overlays from 9200 to 12000; so it will be on top of everything 
(including tooltips from giraffe, which are at 9999)

also bumping the version number.

very minor, small surgical change

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/clockface",
-  "version": "2.11.3",
+  "version": "2.11.4",
   "license": "MIT",
   "main": "dist/index.js",
   "style": "dist/index.css",
--- a/src/Styles/variables.scss
+++ b/src/Styles/variables.scss
@@ -92,7 +92,7 @@ $c-pulsar: #513cc6;
 $cf-z--nav-toggle: 9999;
 $cf-z--draggable-resizer-mask: 9950;
 $cf-z--notifications: 9900;
-$cf-z--portal: 9200;
+$cf-z--portal: 12000;
 $cf-z--nav-menu: 9100;
 $cf-z--nav-mask: 9000;
 $cf-z--tabs-dropdown: 8000;


daviddkkim (changes requested):
This would raise the Z-index of the overlay above the notification. The notification should have higher Z-index of the overlay. 

Is there a rationale for changing the Z index of the overlay as opposed to changing the Z-index of the giraffe components?

jrenee42:
> This would raise the Z-index of the overlay above the notification. The notification should have higher Z-index of the overlay.
> 
> Is there a rationale for changing the Z index of the overlay as opposed to changing the Z-index of the giraffe components?


it actually doesn't raise it.

because the notification is *inside* the portal div; so this pr raises the portal-div stacking context

and the notification itself has a z-index of 9900.  the actual  parts of the overlay just have z-indices of 10,20.30....so on (nth-child *10);

so the notification will still be on top of any overlay components.

this actually helps with notifications, it will make sure they are above any hover legends or annotation legends.
(which are in their own portal (each), with a z-index of 9999; which is above the current portal's z-index of 9200)

the rationale for changing this z-index as it is the simplist fix for making sure that overlays are on top of tooltips.

there are a bunch of numbers in clockface; so trying the simplist one first.

the point of this pr is to get a new clockface version published;
then try out the flaky tests in a sidebranch (with ci, because the flaky tests tend to pass locally but not in the CI system),
and *then* if that works then actually use this version of clockface.

this is a really small tweak that should only improve things and hopefully not break  anything; hence testing in it the CI with the UI first.

if it doesn't work will be easy to revert it.
(passing all my local manual tests, of course)

daviddkkim (approved):
LGTM.

I was wrong about thinking this would mess up the order of the items.
because overlays are within the portal, and notifications appear within the portal as well, it won't break the current stacking we have.

PR #662, opened 2021-08-23
adding onClick on flexbox child so we can pass stop propagation

### LINKED ISSUE #661: Add onClick to flexbox child instead of flexbox
For the accordion, the flexbox has the `onClick` which causes the accordion issues for collapsing and expanding. This will solve that issue if we so can pass things like e.stopPropagation() and have onClick on the flexbox child.

Palakp41:
Closes #661 

### Changes

// Describe what you changed

### Screenshots

### Checklist
Check all that apply

- [ ] Tests pass

### DIFF
--- a/src/Components/FlexBox/FlexBoxChild.tsx
+++ b/src/Components/FlexBox/FlexBoxChild.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {forwardRef} from 'react'
+import React, {forwardRef, MouseEvent} from 'react'
 import classnames from 'classnames'
 
 // Types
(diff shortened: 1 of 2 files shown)


src/Components/FlexBox/FlexBoxChild.tsx
```diff
@@ -48,6 +50,7 @@ export const FlexBoxChild = forwardRef<FlexBoxChildRef, FlexBoxChildProps>(
         style={flexBoxChildStyle}
         data-testid={testID}
         className={flexBoxChildClass}
+        onClick={onClick}
```
hoorayimhelping, line 53: `onClick` can be optional, right? what happens if `onClick` doesn't exist? do we need a no-op function in its place?

daviddkkim (approved):
LGTm

Issue #702, opened 2021-11-09
Dashboards: Hovering over a long graph name causes it to move


hoorayimhelping:
>I think the edit icon thing needs to be absolute positioned to avoid this

https://user-images.githubusercontent.com/146112/140988129-c5c8c761-b6f7-4bae-a33d-f2601cd42139.mov

TCL735:
It is not a long name per se, but rather a name that is long enough to barely fit the top line without the edit icon (pencil) while also too long to fit with the edit icon. When this happens, the line shifts from fitting to not fitting to give room for the edit icon.

ChitlangeSahas:
This is fixed.

PR #715, opened 2022-01-10
fix: resource card editable name pencil css fix

ChitlangeSahas:
Part of https://github.com/influxdata/clockface/issues/702

### Changes

Changes the CSS so that the icon is always aligned with the **first line** of the name.

### Screenshots

Screen Shot 2022-01-10 at 3 54 18 PM

Screen Shot 2022-01-10 at 3 54 08 PM

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/src/Components/ResourceList/Card/ResourceCardEditableName.scss
+++ b/src/Components/ResourceList/Card/ResourceCardEditableName.scss
@@ -19,8 +19,8 @@
   position: relative;
   display: flex;
   flex-wrap: nowrap;
-  align-items: center;
-  max-width: calc(100% - 100px);
+  justify-content: flex-start;
+  max-width: calc(100% - 130px);
 }
 
 /* Ensure placeholder text matches font weight of title */
--- a/src/Components/ResourceList/Card/ResourceCardEditableName.tsx
+++ b/src/Components/ResourceList/Card/ResourceCardEditableName.tsx
@@ -158,7 +158,7 @@ export const ResourceCardEditableName = forwardRef<
             onClick={handleClick}
             data-testid={testID}
           >
-            <span>{name || noNameString}</span>
+            {name || noNameString}
           </span>
         </SpinnerContainer>
         <div
--- a/src/Components/ResourceList/Card/ResourceCardName.scss
+++ b/src/Components/ResourceList/Card/ResourceCardName.scss
@@ -13,10 +13,12 @@
 }
 
 .cf-resource-name--text {
+  display: flex;
+  align-items: center;
+
   font-size: $cf-card--name-font-size;
   font-weight: $cf-card--name-font-weight;
   font-family: $cf-text-font;
-  display: inline-block;
   color: $g17-whisper;
   user-select: none;
 }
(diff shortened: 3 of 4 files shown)


src/Components/ResourceList/Card/ResourceCardEditableName.tsx
```diff
@@ -158,7 +158,7 @@ export const ResourceCardEditableName = forwardRef<
             onClick={handleClick}
             data-testid={testID}
           >
-            <span>{name || noNameString}</span>
```
ChitlangeSahas, line 161: this span was part of the issue.

src/Components/ResourceList/Card/ResourceCardEditableName.scss
```diff
@@ -19,8 +19,8 @@
   position: relative;
   display: flex;
   flex-wrap: nowrap;
-  align-items: center;
```
ChitlangeSahas, line 22: Dont need it in center, rather, at the start of the flex.

hoorayimhelping: one thing to be aware of when modifying CSS is where this class is used. I did a quick poke around the UI (in the chrome console, using `$('.cf-resource-editable-name)` will show you if there are any elements that match that class selector) and found a few places like in paginated tasks and paginated tokens where this class is used. Will justifying the content to `flex-start` mess things up there?

I did notice that changing `align-items: center` changed the vertical placement of text. Do you know what happens when you keep `align-items: center` and add `justify-content: flex-start`?

src/Components/ResourceList/Card/ResourceCardName.scss
```diff
@@ -13,10 +13,12 @@
 }
 
 .cf-resource-name--text {
+  display: flex;
+  align-items: center;
+
   font-size: $cf-card--name-font-size;
   font-weight: $cf-card--name-font-weight;
   font-family: $cf-text-font;
-  display: inline-block;
```
hoorayimhelping, line 19: this selector is also used in more places than dashboards and is something you'll want to take great care changing. Changing from `inline-block` to `flex` might have very odd changes in layout. It might be better to try `inline-flex`, but even then, I'd be very careful changing that, as it might have some strange, unintended consequences beyond just the cards you tested them in.

hoorayimhelping (commented):
Changing a CSS selector is one of the more difficult parts of maintaining CSS because it can sometimes be hard to understand how often it's used. That difficulty is compounded when it's used in a component library like Clockface. Depending on how thoroughly you tested, we may want to tweak this or test more parts of the app.

ChitlangeSahas:
I understand the concerns above, to best navigate the issue of making sure that this doesn't break anything I went through and tested all the pages that have the resource card, I checked: 

- Dashboards,
- Telegraf (under load data)
- Notebooks
- Tasks
- Tokens

Rest of the app has a different flow of renaming, (no pencil, but through the cog icon there is the rename option).

There seem to be no issues. 

cc: @hoorayimhelping

hoorayimhelping (approved):
>I understand the concerns above, to best navigate the issue of making sure that this doesn't break anything I went through and tested all the pages that have the resource card, I checked:
>...
>There seem to be no issues.

outstanding, thank you! Good luck!

PR #716, opened 2022-01-14
fix: fix list uncheck flickering

subirjolly:
## Issue
When you unchecked an item in the list, the entire list would jump to the top. After this change, when you uncheck an item, the list will not show any change other than the actual unchecking action.

## BEFORE
BEFORE

## AFTER
AFTER

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@influxdata/clockface",
-  "version": "3.1.10",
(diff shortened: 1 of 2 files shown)


hoorayimhelping (approved):
love this!!! i hate that scrolling bug

hoorayimhelping (commented):
oh, one thing, can you update the version in package.json please?

PR #717, opened 2022-01-14
fix: fix scrollto positioning issue

subirjolly:
## Issue
As the height of a list item was hard-coded to `24px`, the calculation of `scrollTo` position became out of sync for `xs` because of the extra headers in the list as well as the mismatch of the height of the actual item. I think a good solution would be to use refs for list items but we'd have to take performance into consideration.

For other sizes, the height was still `24` which was absolutely inaccurate and I measured all the sizes and made a map for each of the respective sizes and now the out of sync issue is still there but is very minimal.

## BEFORE
BEFORE

## AFTER
AFTER

### DIFF
--- a/src/Components/List/List.tsx
+++ b/src/Components/List/List.tsx
@@ -168,16 +168,30 @@ const calculateSelectedPosition = (
     return 0
   }
 
-  const itemHeight = 24
+  const heights = {
+    xs: 28,
+    sm: 36,
+    md: 48,
+    lg: 56,
+  }
+
   const items = React.Children.map(children, child =>
     _.get(child, 'props.selected', false)
   )
+
+  const sizes = React.Children.map(children, child =>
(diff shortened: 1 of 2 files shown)


src/Components/List/List.tsx
```diff
@@ -168,16 +168,30 @@ const calculateSelectedPosition = (
     return 0
   }
 
-  const itemHeight = 24
+  const heights = {
+    xs: 28,
+    sm: 36,
+    md: 48,
+    lg: 56,
```
TCL735, line 175: Where do these numbers come from? Is there a common constant for each size?

subirjolly: that's an awesome question. So, I manually did a lookup of the item elements and each of these are sum of height + paddings + margins. I didn't see some common constants for list items.

hoorayimhelping (approved):

PR #724, opened 2022-02-07
feat: add new component Creatable Type Ahead Dropdown

appletreeisyellow:
Closes influxdata/ui#3680

This PR created a new component called "Creatable Type Ahead Dropdown," a more flexible version of "Type Ahead Dropdown." 

Here are the features provided with this new component:
- Type ahead search in dropdown menu
- Accept typed values that are not in the dropdown menu
- Customizable dropdown items
- Optional color preview

### Changes

- Add a new component `CreatableTypeAheadDropdown`
- Add this new component into storybook with examples
- Add `colorPreview` prop to `Input` component
- Move `ColorPreview` component and its style into a single file

### Screenshots

https://user-images.githubusercontent.com/14298407/152875227-c79a1825-d8e4-4286-9764-c68b63da5899.mov

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/src/Components/ColorPicker/ColorPicker.scss
+++ b/src/Components/ColorPicker/ColorPicker.scss
@@ -94,22 +94,3 @@ $color-picker--margin: 0;
     padding-left: $cf-marg-d - $cf-marg-a;
   }
 }
-
-.cf-color-picker--selected {
-  pointer-events: none;
-  z-index: 2;
-  position: absolute;
-  top: 50%;
-  left: $cf-marg-c;
-  transform: translate(-50%, -50%);
-  width: 18px;
-  height: 18px;
-  border-radius: 50%;
-  border: $cf-border solid $g5-pepper;
-  transition: background-color $cf-transition-default,
-    border-color $cf-transition-default;
-
-  .cf-input:hover + & {
-    border-color: $g7-graphite;
-  }
-}
--- a/src/Components/ColorPicker/ColorPreview.tsx
+++ b/src/Components/ColorPicker/ColorPreview.tsx
@@ -9,6 +9,7 @@ import {StandardFunctionProps} from '../../Types'
 import './ColorPreview.scss'
 
 interface ColorPreviewProps extends StandardFunctionProps {
+  /** Any CSS color value is good */
   color: string
 }
 
--- a/src/Components/ColorPicker/ColorPreview.tsx
+++ b/src/Components/ColorPicker/ColorPreview.tsx
@@ -9,7 +9,6 @@ import {StandardFunctionProps} from '../../Types'
 import './ColorPreview.scss'
 
 interface ColorPreviewProps extends StandardFunctionProps {
-  /** Hex color in #000000 format */
   color: string
 }
 
--- a/src/Components/ColorPicker/ColorPreview.tsx
+++ b/src/Components/ColorPicker/ColorPreview.tsx
@@ -1,0 +1,27 @@
+// Libraries
+import React, {FunctionComponent} from 'react'
+import classnames from 'classnames'
+
+// Types
+import {StandardFunctionProps} from '../../Types'
+
+// Styles
+import './ColorPreview.scss'
+
+interface ColorPreviewProps extends StandardFunctionProps {
+  /** Hex color in #000000 format */
+  color: string
+}
+
+export const ColorPreview: FunctionComponent<ColorPreviewProps> = ({
+  color,
+  className,
+}) => {
+  const colorPreviewClass = classnames('cf-color-preview', {
+    [`${className}`]: className,
+  })
+
+  return <div className={colorPreviewClass} style={{backgroundColor: color}} />
+}
+
+ColorPreview.displayName = 'ColorPickerPreview'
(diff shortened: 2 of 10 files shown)


src/Components/ColorPicker/ColorPicker.scss
```diff
@@ -94,22 +94,3 @@ $color-picker--margin: 0;
...
-  height: 18px;
-  border-radius: 50%;
-  border: $cf-border solid $g5-pepper;
-  transition: background-color $cf-transition-default,
-    border-color $cf-transition-default;
-
-  .cf-input:hover + & {
-    border-color: $g7-graphite;
-  }
-}
```
hoorayimhelping, line 115: does this removal affect the existing color picker?

src/Components/ColorPicker/ColorPreview.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+// Types
+import {StandardFunctionProps} from '../../Types'
+
+// Styles
+import './ColorPreview.scss'
+
+interface ColorPreviewProps extends StandardFunctionProps {
+  /** Hex color in #000000 format */
+  color: string
```
hoorayimhelping, line 13: style suggestion: this comment might become unnecessary and the code might be more self-documenting if the prop was called `hexColor` or `colorHex`, but I don't know all the details around this prop needing to be named `color`

src/Components/Dropdowns/Composed/CreatableTypeAheadDropdown.tsx
```diff
@@ -0,0 +1,222 @@
...
+   * The color is determined by the selected or typed option in #000000 format.
+   * If both icon and this props are set, icon will take priority */
+  inputColorPreviewOn?: boolean
+  /** Optional theme of menu */
+  menuTheme?: DropdownMenuTheme
+  /** Optional maximum pixel height menu */
+  menuMaxHeight?: number
+  /** Optional customized dropdown item */
+  customizedDropdownItem?: (displayText: string) => JSX.Element
+}
```
hoorayimhelping, line 50: i know that this was copied over from the existing `TypeaheadDropdown` but for the future: in general, comments that restate the obvious aren't necessary and can actually make code harder to read. 

examples of comments stating the obvious:
```ts
 /** Fires when an option is clicked, used to update state */
onSelect: (option: string) => void

 /** Optional status of input */
inputStatus?: ComponentStatus

/** Optional placeholder text when no option is selected */
placeholder?: string

/** Optional theme of menu */
menuTheme?: DropdownMenuTheme
```

i think this:
```ts
  /** Optional color preview to be displayed to the left of text.
   * The color is determined by the selected or typed option in #000000 format.
   * If both icon and this props are set, icon will take priority */
```
is an example of a good comment. it provides context that may not be obvious from reading the variable name :+1:

src/Components/Inputs/Input.tsx
```diff
@@ -247,7 +257,8 @@ export const Input = forwardRef<InputRef, InputProps>(
         />
         {clearElement}
         {type === InputType.Checkbox && <div className={inputCheckboxClass} />}
-        {iconElement}
+        {/** If both icon and colorPreview are set in props, icon has higher priority */
+        !!icon ? iconElement : colorElement}
```
hoorayimhelping, line 261: is `colorElement` always going to exist? this kind of implies that if `icon` is falsy, `colorElement` will always be shown. In reality, if they both don't exist, nothing will get rendered. But the intent of the code isn't as clear - this makes it seem like if `icon` is falsy, `colorElement` should be always shown.

it might be clearer to do something like this above where `iconElement` and `colorElement` are defined:

```ts
let iconElement = null
if (icon) {
  iconElement = <Icon glyph={icon} className="cf-input-icon" />
} else if (colorPreview) {
  iconElement = <ColorPreview color={colorPreview} className="cf-input-icon" />
}
```

then down here, leaving this as is
```ts
{iconElement}
```

that way, `icon` takes precedence, and the color preview is only shown if icon doesn't exist and `colorPreview` exists.

The outcome will be the same, but the intent of the code is a little bit clearer - that icon takes precedence, and the color preview is only shown when icon doesn't exist and `colorPreview` does exist

hoorayimhelping (approved):
LGTM - have a couple of small comments but none of them are blocking. You can take them or leave them

src/Components/ColorPicker/ColorPreview.tsx
appletreeisyellow: I looked more into the code and found that the prop is not limited to hex color., so I updated the comment

src/Components/Inputs/Input.tsx
appletreeisyellow: Thanks! This suggestion makes more sense!

src/Components/ColorPicker/ColorPicker.scss
appletreeisyellow: This removal doesn't affect the existing color picker even though this portion of the code is moved to a separate file "src/Components/ColorPicker/ColorPreview.scss." I also double checked by looking into Google chrome dev tool. When the `cf-input` element is on hover, the border-color for color preview does change.

PR #740, opened 2022-03-10
fix: Added Pagination Input Capability

brandenTenbrink:
Closes: https://github.com/influxdata/ui/issues/3918

### Changes

An input component was added to the existing Pagination Component. It is an optional component that can be enabled if the user has a large # of pages available. 

Input handles any number within 1 - TotalPages.

### Screenshots

Screen Shot 2022-03-10 at 4 56 53 PM

### Checklist
Check all that apply

- [x] Tests pass


hoorayimhelping (approved):
Looks good to me. 

So as I understand it, `PaginationInput` is a component that can't stand alone, its rendering is handled by `PaginationNav`, right?

Before merging this in, can you bump the version in `package.json` to `3.5.0` since this is new, non-breaking functionality and update the changelog.md file please? The reason we like to update the version in package in the same PR that the code is changed (when possible) is so that on the [releases](https://github.com/influxdata/clockface/releases) page, the git commit points to the functionality change, rather than the PR to bump package. (`v3.3.0` points to the correct code change, `v3.4.0` pointing to a PR that bumps package.json)

brandenTenbrink:
> 

That is correct. It can be turned on and off as needed, but cannot be rendered without a PaginationNav. 
And will do!

PR #755, opened 2022-04-08
Update Clockface Fonts

brandenTenbrink:
Closes #

### Changes

Updated the Clockface fonts:

Code Font: IBMPlexMono => Roboto Mono

Text Font: Rubik => ProximaNova

### Screenshots

Screen Shot 2022-04-08 at 1 28 35 AM

Screen Shot 2022-04-08 at 1 28 43 AM

### Checklist
Check all that apply

- [x] Tests pass


brandenTenbrink:
@hoorayimhelping Figured we could try and avoid a meeting if you just wanna check the webpack stuff here async. I think its right since everything is loading fine.

hoorayimhelping (commented):
Looks good to me, not counting the consideration on versioning. The webpack changes look fine - doesn't seem like a big deal, just adding in support for more font formats.

A different consideration: should this be considered a breaking change? Would a hypothetical (or actual) consumer of the Clockface library need to change their application in response to pulling in this change? Could this change potentially break an application using Clockface as its component library? @mavarius spoke to me about needing to QA these changes before moving them to prod. To me that implies that things could break if a user pulls in these changes; I think that implies we should bump Clockface up to version `4.0.0` with this change.

[Semver](https://semver.org/) (which is the `x.y.z` versioning scheme we use in libraries) for reference. This is one of those cases where it'll take some human judgement to determine what version this is, but I'm leaning towards bumping Clockface up to `4.0.0` since changes in fonts and headings might cause cascading CSS changes in an application.

brandenTenbrink:
> Looks good to me, not counting the consideration on versioning. The webpack changes look fine - doesn't seem like a big deal, just adding in support for more font formats.
> 
> A consideration: should this count as a breaking change? Would a hypothetical (or actual) consumer of the Clockface library need to change their application in response to pulling in this change? Could this change potentially break an application using Clockface as its component library? @mavarius spoke to me about needing to QA these changes before moving them to prod. To me that implies that things could break if a user pulls in these changes; I think that implies we should bump Clockface up to version `4.0.0` with this change.
> 
> [Semver](https://semver.org/) (which is the `x.y.z` versioning scheme we use in libraries) for reference. This is one of those cases where it'll take some human judgement to determine what version this is, but I'm leaning towards bumping Clockface up to `4.0.0` since changes in fonts and headings might cause cascading CSS changes in an application.

Yeah, I agree on 4.0.0 for this change. There are several updates going out between now and next week, so I will be bumping the version once I know their order.

mavarius (approved):
Looks good!

Issue #753, opened 2022-04-07
Subway Nav Additions


helenosheaa:
**Describe the solution you'd like**
The subway nav currently has "Setting Up" hardcoded as the title text. I need it to be handed in as a prop so we can match the cloud subscriptions UI where it sometimes has the name of a subscription in that location.

I also need the icon ticks not to be rendered in certain situations. We could had in optional prop to the steps that says whether to turn the icon into a checkmark upon completion. This is for the edit view of the subscriptions design.

Example of different text:
<img width="510" alt="Screen Shot 2022-04-07 at 12 50 04 PM" src="https://user-images.githubusercontent.com/38860767/162255801-63698042-c11d-4edb-aa4a-389d6ddbdd3f.png">

Should be able to progress to next section without the top item becoming a tick:
<img width="222" alt="Screen Shot 2022-04-07 at 12 47 57 PM" src="https://user-images.githubusercontent.com/38860767/162255939-5cd1eade-b801-4e85-ac85-73cfd14ed2df.png">

PR #759, opened 2022-04-13
feat: subway nav additions

helenosheaa:
Closes https://github.com/influxdata/clockface/issues/753
Helps with https://github.com/influxdata/data-acquisition/issues/240 

### Changes

I've added two optional props: settingUpHeader which defaults to "Setting Up", if a prop is not handed in. And showCheckmark which defaults to true unless it's handed in.

Both these changes are things which are needed only in the subscriptions edit view. Below see video of the changes to the edit view, the create view still looks the same and has no additional props needing to be handed in. 

No checkmarks on progress through the single page edit view & subscription name in place of setting up text.

### Screenshots

https://user-images.githubusercontent.com/38860767/163233086-3dc86632-5ac6-4ba6-bb74-d9893f014f3a.mov

@hoorayimhelping has agreed to publish once this has been approved and I'll swap it out in the UI.

### Checklist
Check all that apply

- [x] Tests pass


src/Components/SubwayNav/SubwayNav.tsx
```diff
@@ -19,6 +19,8 @@ interface OwnProps {
   settingUpIcon: JSX.Element
   settingUpText: string
   setupTime?: string
+  settingUpHeader?: string
+  hideCheckmark?: boolean
```
hoorayimhelping, line 23: I think we should consider renaming this to `showCheckMark` so that we don't introduce double negatives with `!hideCheckMark`.

That would change the code calling `SubwayNavStep` to something like

```ts
text={value.name}
showCheckmark={this.props.showCheckmark !== false}
```

and the code in `SubwayNavStep` to something like

```ts
stepIsComplete && showCheckmark 
```

what do you think?

hoorayimhelping (commented):
I left a comment that's a suggestion - it's not a hard requirement, just something I think might make the code slightly easier to follow. What are your thoughts on that?

can you also bump the version in `package.json` to `3.11.0` and update `CHANGELOG.md` with a link to this PR please?

src/Components/SubwayNav/SubwayNav.tsx
helenosheaa: makes sense to me, i've updated it :)

helenosheaa:
thanks for the speedy review, I've updated the version number, changelog and made your suggested change.

brandenTenbrink:
Everything looks fine to me, my only suggestion is would we want to use a defaultProps to set "Setting up" instead of having it done inline code. Makes it more readable imo.

Screen Shot 2022-04-18 at 12 19 10 PM

@helenosheaa @hoorayimhelping

helenosheaa:
@brandenTenbrink, do we have a practice of adding static default props like that in clockface? I'm a fan of an inline operator in something simple like this, but not against changing it.

@hoorayimhelping thoughts?

hoorayimhelping:
@helenosheaa @brandenTenbrink I prefer to use default props as well. I find it generally conceptually simpler to comprhened when the default value of a property is set in a single place. If a prop is used in multiple places, the default will have to be set multiple times.

But at the same time, we don't use `defaultProps` anywhere in Clockface, so I don't think we should block this PR on it, unless Branden feels _very_ strongly about it. If it's more of Branden's preferred way of doing things and not a hard requirement, we can work together to set a standard for it going forward, and replace inline uses.

hoorayimhelping (approved):

PR #793, opened 2022-06-30
refactor: MenuDropdown Cleanup

ChitlangeSahas:
There were a few unnecessary variables that needed to be removed and some renamed from generic names to more specific ones so it's easier to follow in the future. :)

### Changes

// Describe what you changed

### Checklist
Check all that apply

- [x] Tests pass

### DIFF
--- a/src/Components/Dropdowns/Composed/MenuDropdown.tsx
+++ b/src/Components/Dropdowns/Composed/MenuDropdown.tsx
@@ -326,7 +326,7 @@ export const MenuDropdown: FC<MenuDropdownProps> = ({
               itemCount={queryResults.length}
               itemSize={50}
               width={'100%'}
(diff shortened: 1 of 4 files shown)


src/Components/Dropdowns/Composed/MenuDropdown.tsx
```diff
@@ -33,7 +33,7 @@ export interface MenuItem {
 }
 
 export interface SubMenuItem {
-  id: string
+  id: string | number
```
wdoconnell, line 36: Thanks!  FYI for others: this is the one that's crucial for the multi-org work. Quartz returns arrays of `Accounts` with an id of type `number`, so we'd like to be able to us those as `subMenuItems` without having to re-map in the UI.

PR #794, opened 2022-07-01
fix: Multi-org dropdown shouldn't show typeahead if only one account/org

wdoconnell:
Per [UI 4047](https://github.com/influxdata/ui/issues/4047) and [UI 4051](https://github.com/influxdata/ui/issues/4051), the multi-org dropdown:

1. Should not display the 'switch account' or 'switch org' menu if there's only one account or org.
2. Should check for the currently 'selected' account or org's `id` after converting to a string: Org arrays have `string` ids, but account arrays have `number` ids.

*EDIT*: Removed prior change involving sorting selected item to first part of the array. That is easier to handle in a sort in the UI.
 
### Checklist
Check all that apply

- [x] Updated documentation to reflect changes [N/A]
- [X] Added entry to top of Changelog with link to PR (not issue)
- [X] Tests pass
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed) [N/A - InfluxData employee)

### DIFF
--- a/src/Components/Dropdowns/Composed/MenuDropdown.tsx
+++ b/src/Components/Dropdowns/Composed/MenuDropdown.tsx
@@ -346,7 +346,10 @@ export const MenuDropdown: FC<MenuDropdownProps> = ({
                       id={value.id.toString()}
                       value={value}
                       onClick={() => doSelection(value, true)}
-                      selected={value.id === selectedItem?.id}
+                      /* Values need to be compared as string because account items have number ids*/
+                      selected={
+                        value.id.toString() === selectedItem?.id.toString()
+                      }
                       className={classN}
                       trailingIconOnSelected={true}
                     >
--- a/src/Components/Dropdowns/Composed/MenuDropdown.tsx
+++ b/src/Components/Dropdowns/Composed/MenuDropdown.tsx
@@ -39,7 +39,7 @@ export interface SubMenuItem {
 
 export interface MenuDropdownProps extends StandardFunctionProps {
   /** A default pre-selected Option */
-  selectedOption: SubMenuItem
+  selectedOption?: SubMenuItem
   /** List of href options to render in the main menu */
   options: MenuItem[]
   /** List of options to render in the sub type-ahead menu */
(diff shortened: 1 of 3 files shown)


src/Components/Dropdowns/Composed/MenuDropdown.tsx
```diff
@@ -329,10 +339,11 @@ export const MenuDropdown: FC<MenuDropdownProps> = ({
           <List
             className="List"
             height={150}
-            itemCount={queryResults.length}
+            itemCount={queryResults.length + 1}
             itemSize={50}
             width={menuWidth}
-            itemData={queryResults}
+            /* Multi-org UI tickets #4051 and #4047: display the currently selected item first. */
+            itemData={[selectedOption, ...queryResults]}
```
ChitlangeSahas, line 346: instead of overwriting the list, we can use the `scrollToItem` property of this component. 

https://react-window.vercel.app/#/api/FixedSizeList:~:text=scrollToItem(index%3A%20number%2C%20align%3A%20string%20%3D%20%22auto%22)%3A%20void

ChitlangeSahas: What do you think about that?

wdoconnell: Hm - so, that's an option, but since it would mean the item isn't displaying at the top, we should ask @Luckette . The reason I've done it this way is so the item is always at the top - it's pulled out of the array during the filter process, then pushed back in as item `[0]` when displayed to the user.

"The currently selected account should always display at the top of the dropdown regardless of the ABC order so that the user can quickly see which account is currently selected."

ChitlangeSahas: Okay okay that makes sense, thank you for explaining :)

ChitlangeSahas (approved):

src/Components/Dropdowns/Composed/MenuDropdown.tsx
wdoconnell: @ChitlangeSahas I thought about this a little more, and figured out a better way to do it in the same `sort` the UI is already doing - so removed this change from clockface.

ChitlangeSahas: Thanks for updating me of the change !

src/Components/Dropdowns/Composed/MenuDropdown.tsx
```diff
@@ -341,7 +346,10 @@ export const MenuDropdown: FC<MenuDropdownProps> = ({
                       id={value.id.toString()}
                       value={value}
                       onClick={() => doSelection(value, true)}
-                      selected={value.id === selectedItem?.id}
+                      /* Values need to be compared as string because account items have number ids*/
+                      selected={
+                        value.id.toString() === selectedItem?.id.toString()
```
ChitlangeSahas, line 351: 👍

Issue #834, opened 2022-09-16
DraggableResizer collapse doesn't work


appletreeisyellow:
**Describe the bug**

The `DraggableResizer` doesn't seem to collapse correctly when there are odd number of panels

**To Reproduce**
1. Go to `DraggableResizer.stories.tsx` the example with `4 Panels`
2. Remove the forth panel and update default `positions` to be `[0.25, 0.8]`
3. Remove `isCollapsible={true}` for the second panel and add it to the third panel
4. Run story locally `npm run storybook`

**Expected behavior**
The third panel should collpse

**Screenshots**

https://user-images.githubusercontent.com/14298407/190728718-f33558d5-3603-4597-898c-62734f8d86c0.mov

**Browser**
 - OS: Mac OS
 - Browser: Chrome
 - Version: 105

PR #840, opened 2022-09-21
fix: Odd Panels would not Collapse properly

brandenTenbrink:
Closes https://github.com/influxdata/clockface/issues/834

### Changes

Fixed a bug that prevented the last odd panel from collapsing in the 1 (right or down) direction.
Created a proper offset value for odd and even numbered panels. 

Also updated storybook to have 1 example of even and odd panels.

### Screenshots

### Checklist
Check all that apply

- [ ] Tests pass

### DIFF
--- a/src/Components/DraggableResizer/DraggableResizer.tsx
+++ b/src/Components/DraggableResizer/DraggableResizer.tsx
@@ -182,7 +182,6 @@ export const DraggableResizerRoot: FunctionComponent<DraggableResizerProps> = ({
         newPositions[i] = newPos
       }
     }
-    console.log('output', newPositions)
     onChangePositions(newPositions)
   }
 
--- a/src/Components/DraggableResizer/DraggableResizer.tsx
+++ b/src/Components/DraggableResizer/DraggableResizer.tsx
@@ -48,6 +48,8 @@ export const DraggableResizerRoot: FunctionComponent<DraggableResizerProps> = ({
   backgroundStyle,
   handleBarStyle,
 }) => {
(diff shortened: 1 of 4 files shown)


src/Components/DraggableResizer/DraggableResizer.tsx
```diff
@@ -131,7 +133,9 @@ export const DraggableResizerRoot: FunctionComponent<DraggableResizerProps> = ({
 
   const calculateCollapsePosition = (direction: number, dragIndex: number) => {
     const totalDrag = handlePositions.length - 1
-    const directionIndex = direction === 0 ? dragIndex : dragIndex + 2
+    const directionOffset = PANEL_COUNT % 2 == 0 ? 2 : 3
```
appletreeisyellow, line 136: Just wondering, what does `2` and `3` mean here?

appletreeisyellow (approved):
LGTM! Thanks for fixing this 🙌 

Just a question below, not a blocker

src/Components/DraggableResizer/DraggableResizer.tsx
brandenTenbrink: It's an offset on top of the drag index.
So the index is 0 - X depending on the number of handles. We add this offset to make it 0 - X + 2 if its even or 0 -X + 3 if its odd. 

This summed value is then multiplied against 0.25 (which is each individual breakpoint for the panels). And that's what pushes it closer to the 0 or 1 value that is needed to collapse it.

This component is extremely complicated hah

PR #844, opened 2022-09-26
fix: Icons Rendering Wrong Font Files

brandenTenbrink:
Closes #

### Changes

Updated the font files so the correct icons are now being rendered in all cases.

Refreshes all the icon files and scss, so it is correct

Example: The PieChart Icon was rendering the Save Outline icon.
Save Icon was rendering the Piechart

### Screenshots

### Checklist
Check all that apply

- [ ] Tests pass


src/Styles/icon.scss
```diff
@@ -61,7 +64,7 @@
   &.CollapseRight:before {
     content: '\e94d';
   }
-  &.DoubleCaretVertical:before {
+  &.ChangeState:before {
```
ChitlangeSahas, line 67: This will break multi-org, we use `DoubleCaretVertical` name in UI.

ChitlangeSahas: FYI @wdoconnell , please correct me if I am misunderstanding.

brandenTenbrink: Uhh, it shouldn't have removed anything.

brandenTenbrink: Unless this is an icon not in our icomoon set.

ChitlangeSahas: Correct, but if something was named `DoubleCaretVertical` i.e referred as `IconFont.DoubleCaretVertical`,

and now it's called ChangeState, Ui wont find it.

brandenTenbrink: Yup, looks like theres a mismach between naming. I'll update that in icomoon.

ChitlangeSahas: Thank you.

src/Styles/icon.scss
```diff
@@ -29,29 +29,32 @@
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
   &.FolderOpen:before {
-    content: "\e956";
+    content: '\e956';
   }
-  &.Bill:before {
-    content: '\e954';
+  &.Timer:before {
```
ChitlangeSahas, line 34: Seems like this was just cut and pasted from the bottom?

brandenTenbrink: Yeah some of them got moved, since I refreshed teh entire file from the list generated by icomoon

ChitlangeSahas: Just trying to avoid unnecessary diff, if that is the case. This makes it easier to avoid bugs, specially the ones where tests can't catch them. (like this case)

brandenTenbrink: We should let this one slide, then ensure we follow icomoon's order moving forward then. I must have pasted these on top in the past.

hoorayimhelping: It looks like these used to be alphabetized. We should maintain that ordering if there's no other ordering present

hoorayimhelping: I agree with Sahas here. The changes in ordering give the impression that the diff is bigger than it is, and hides the actual change

ChitlangeSahas: I would be okay with letting it slide if we hadn't had bad previous experiences with shipping changes that didn't feel like a big deal, but then they clog up the pipeline for days. 

Probably just paranoia but the consequences in the past have been very real 😄

ChitlangeSahas:
Looks good to me, just let me know when you fix the diff, and I can approve. 👍

brandenTenbrink:
> Looks good to me, just let me know when you fix the diff, and I can approve. 👍

done

ChitlangeSahas (approved):
There you go, much better to understand that you're just changing `content` 's value. :)

Issue #838, opened 2022-09-19
Clockface 4.0: Dropdowns [in progress]


taramk:
* Dropdown
* DropdownButton
* DropdownDivider
* DropdownItem
* DropdownItemEmpty
* DropdownLinkItem
* DropdownMenu
* SelectDropdown
* TypeAheadDropdown
* CreateableTypeAheadDropdown
* MultiSelectdropdown

ChitlangeSahas:
Part of [#6090](https://github.com/influxdata/ui/issues/6090)

PR #876, opened 2022-10-18
feat(Dropdowns): Clockface 4 dropdown updates

ChitlangeSahas:
Closes https://github.com/influxdata/clockface/issues/838

Couple screenshots:

`SelectDropdown`:

Screen Shot 2022-10-17 at 9 22 16 PM

`TypeAheadDropdown`:

Screen Shot 2022-10-17 at 9 27 58 PM

`CreateableTypeAheadDropdown`: 

Screen Shot 2022-10-17 at 9 28 43 PM

`MultiSelect`:

Screen Shot 2022-10-17 at 9 29 16 PM

- [ ] Tests pass

### DIFF
--- a/src/Components/Button/Button.scss
+++ b/src/Components/Button/Button.scss
@@ -222,7 +222,12 @@
 
 .cf-button-primary,
 .cf-button-success {
-  @include buttonColorModifier($cf-turquoise, $cf-turquoise, $cf-grey-1, $cf-grey-1);
+  @include buttonColorModifier(
+    $cf-turquoise,
+    $cf-turquoise,
+    $cf-grey-1,
+    $cf-grey-1
+  );
 }
 
 .cf-button-tertiary {
@@ -237,10 +242,15 @@
 
 .cf-button-warning,
 .cf-button-danger {
-  @include buttonColorModifier($cf-carnation, $cf-carnation, $cf-grey-1, $cf-grey-1);
+  @include buttonColorModifier(
+    $cf-carnation,
+    $cf-carnation,
+    $cf-grey-1,
+    $cf-grey-1
+  );
 
-  &:focus{
-    @include focus-shadow($cf-carnation)
+  &:focus {
+    @include focus-shadow($cf-carnation);
   }
 }
 
--- a/src/Components/Dropdowns/Dropdown.scss
+++ b/src/Components/Dropdowns/Dropdown.scss
@@ -43,6 +43,7 @@ $dropdown-item--padding-v: $cf-space-2xs;
   border-radius: $cf-radius;
   box-shadow: 0 2px 5px 0.6px rgba($g0-obsidian, 0.2);
   padding: $cf-space-2xs;
+  border: rgba(255, 255, 255, 0.05) solid $cf-border;
 }
 
 .cf-dropdown-menu--contents {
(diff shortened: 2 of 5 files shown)


src/Components/Dropdowns/Dropdown.scss
```diff
@@ -52,6 +53,8 @@ $dropdown-item--padding-v: $cf-space-2xs;
   flex-direction: column;
   align-items: stretch;
   cursor: pointer;
+  margin-right: 2px;
+  margin-left: 1px;
```
ChitlangeSahas, line 57: These two stick out, but they're just to please the `absolute` positioned, Scrollbar library we're using

src/Components/Button/Button.scss
```diff
@@ -222,7 +222,12 @@
 
 .cf-button-primary,
 .cf-button-success {
-  @include buttonColorModifier($cf-turquoise, $cf-turquoise, $cf-grey-1, $cf-grey-1);
+  @include buttonColorModifier(
+    $cf-turquoise,
+    $cf-turquoise,
+    $cf-grey-1,
+    $cf-grey-1
+  );
```
ChitlangeSahas, line 230: Just ran prettier, hence the diff.

src/Components/Dropdowns/Dropdown.scss
hoorayimhelping: Can this be solved using `position` and `left` and `right` rather than with using margins? Semantically, there's a big difference between position and margin, and we should use positioning for laying things out, and margins for when we need speec around an element. I'm not super familiar with how this is used, though

ChitlangeSahas: Got you! I'll try that!

ChitlangeSahas: Screen Shot 2022-10-18 at 8 19 39 AM

It still cuts off the right end of it, I think this is what is happening:

IMG_D4759C8952C6-1

So, when I add the margin, the border falls under the available space.

Unfortunately, I am not able to change the width of the inner wrapper, since it's controlled by the react-custom-scrollbar library.

ChitlangeSahas: This was always a problem, but we never noticed because we didn't use `borders` as an indicator for being selected. Now that we use the border, we see that the 1px of the right is cut off.

src/Components/Dropdowns/Dropdown.scss
```diff
@@ -69,10 +72,18 @@ $dropdown-item--padding-v: $cf-space-2xs;
 .cf-dropdown-divider {
   @include dropdownItemStyles();
   font-weight: $cf-font-weight--medium;
+  text-transform: uppercase;
+  color: #88889b;
```
wdoconnell, line 76: Since this is just a divider - does `text-transform: uppercase` have any effect?

wdoconnell: Gotcha - thanks!

wdoconnell (approved):

hoorayimhelping (approved):

src/Components/Dropdowns/Dropdown.scss
taramk: is it causing problems to have the border? julia and i also discussed maybe not needing a background color at all on selected items. we included it because that's what we have now—adjusted to be lower contrast so it's less distracting—but i think the selected state for the radio buttons and checkboxes has enough contrast that we might not need the background color+border as well. let me know what you think. @ChitlangeSahas @hoorayimhelping @juliajames12

PR #886, opened 2022-11-03
chore(Clockface 4): remove unused color variables

ChitlangeSahas:
A chore to remove unused color variables. More cleanup coming in the future, we need to map the old colors to the new ones, this is the first step.

### DIFF
--- a/src/Styles/variables.scss
+++ b/src/Styles/variables.scss
@@ -25,12 +25,7 @@ $cf-lime-green: #51d91c;
 $cf-carnation: #e65b5b;
 
 // Background Colors
-$cf-elephant: #1e3a46;
 $cf-martinique: #332548;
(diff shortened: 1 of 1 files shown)


TCL735:
`$c-topaz: #e85b1c;` and `$c-pulsar: #513cc6;` are still used in Giraffe. Do we need to replace and remove them there?

TCL735 (approved):

ChitlangeSahas:
> `$c-topaz: #e85b1c;` and `$c-pulsar: #513cc6;` are still used in Giraffe. Do we need to replace and remove them there?

Yes! I'm thinking this is a good opportunity to replace all old colors, what happened with CF 3 is we only replaced surface level stuff, while keeping the old stuff. This made it so that there are a lot of places the old colors are still used.

